Issue #29, opened 2021-02-12
Inconsistencies in units of energy in models.py


nuberoi:
The method `get_initialspectra` inside the model classes uses seems to apply the log function to an `astropy.Quantity` term, leading to a `UnitTypeError` and a `UnitConversionError` when the notebook `Flavor_XForm.ipynb` is run.

sybenzvi:
I can't reproduce this issue, and I am definitely using the current checkout of the main branch. You may need to double-check your user environment @nuberoi... something suspicious is going on.

nuberoi:
Exact error message:

---------------------------------------------------------------------------
UnitConversionError                       Traceback (most recent call last)
c:\programdata\anaconda3\lib\site-packages\astropy\units\quantity_helper\helpers.py in get_converter(from_unit, to_unit)
     31     try:
---> 32         scale = from_unit._to(to_unit)
     33     except UnitsError:

c:\programdata\anaconda3\lib\site-packages\astropy\units\core.py in _to(self, other)
    950         raise UnitConversionError(
--> 951             f"'{self!r}' is not a scaled version of '{other!r}'")
    952 

UnitConversionError: 'Unit("erg / s")' is not a scaled version of 'Unit(dimensionless)'

During handling of the above exception, another exception occurred:

UnitConversionError                       Traceback (most recent call last)
c:\programdata\anaconda3\lib\site-packages\astropy\units\quantity_helper\helpers.py in helper_dimensionless_to_dimensionless(f, unit)
    146     try:
--> 147         return ([get_converter(unit, dimensionless_unscaled)],
    148                 dimensionless_unscaled)

c:\programdata\anaconda3\lib\site-packages\astropy\units\quantity_helper\helpers.py in get_converter(from_unit, to_unit)
     34         return from_unit._apply_equivalencies(
---> 35                 from_unit, to_unit, get_current_unit_registry().equivalencies)
     36     except AttributeError:

c:\programdata\anaconda3\lib\site-packages\astropy\units\core.py in _apply_equivalencies(self, unit, other, equivalencies)
    887             "{} and {} are not convertible".format(
--> 888                 unit_str, other_str))
    889 

UnitConversionError: 'erg / s' (power) and '' (dimensionless) are not convertible

During handling of the above exception, another exception occurred:

UnitTypeError                             Traceback (most recent call last)
<ipython-input-8-3405dee786df> in <module>
----> 1 fig = plot_spec(model, NoTransformation)

[...]

PR #30, opened 2021-02-12
Removed units from L and Ea in get_initialspectra as a potential fix to Issue #29

nuberoi:


sybenzvi (commented):
It looks like on my system, running `np.interp` on to astropy.Quantity arrays returns a type `numpy.float64`, so this fix actually breaks my notebooks. Instead I will submit a different fix that checks the type and removes the units if necessary.

PR #34, opened 2021-02-26
Make NeutrinoDecay Calculations more Efficient

nuberoi:
Remove for loop that computes neutrino decay probabilities


sybenzvi:
I haven't checked this calculation in forever... no idea if this is still relevant.

PR #38, opened 2021-04-28
Scripts changed to provide the neutrino lightcurves as output

mcolomermolla:
Modification to the scripts:

- to_snowglobes: in case dt of nbins is not provided, the binning is fixed to 1 ms, and number of bins given by tmin and tmax
- from_snowglobes:
~ Function returns now the integrated number of events over energy and summing all interaction channels, for each time bin used in the lightcurve
~ In case only one detector option is chosen, the script outputs a txt file with first column the edge of the time bin and second the number of events in the time bin, only for the smeared+weighted case, to build the detected lighturve

### DIFF
--- a/snewpy/from_snowglobes.py
+++ b/snewpy/from_snowglobes.py
@@ -3,8 +3,8 @@ from __future__ import unicode_literals
 import os
 import tarfile
 import zipfile
-import matplotlib as mpl
-mpl.use('Agg')
+#import matplotlib as mpl
+#mpl.use('Agg')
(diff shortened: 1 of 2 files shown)


JostMigenda:
As far as I understand, this is effectively superseded by #67, right?

Additionally, this PR is targeted at a stale branch, as discussed in #44. So I think we can close this PR (and, if necessary, open a new one with any changes not included in #67 and target that at the main branch).

mcolomermolla:
Yes, this PR can definitely be closed now! :)

PR #43, opened 2021-07-15
run unit tests automatically

JostMigenda:
@sybenzvi added a basic GitHub Action to run unit tests last year, but thus far, it has not been running. This PR fixes that and also ensures the tests run under more recent Python versions.

Note that the tests [are currently failing](https://github.com/SNEWS2/snewpy/runs/3077655323). @sgriswol, Segev said you’re currently working on fixing/adding the tests?


sybenzvi:
Looks like the tests are failing because h5py is not included as a dependency.

JostMigenda:
… and because of an import error from snewpy.flavor_transformation; yes. I’ve fixed both in the commit just now, but am getting two new assertion failures:

In `test_Nakazato_vanilla()`:
```
>       assert(np.interp(t.to(tunit), model.time, model.meanE[Flavor.NU_E])*Eunit == 6.79147181061522*u.MeV)
E       assert (<Quantity 6.79147181 MeV> * Unit("MeV")) == (6.79147181061522 * Unit("MeV"))
```
This looks like just a rounding error; we can probably change that to `assert(abs(…) < 1e-8)`.

And in `test_2fd()`:
```
>       assert(xform.prob_ee(t, E) == 0.5)
E       assert <Quantity 0.15956689> == 0.5
```
This looks like the test and the implementation of TwoFlavorDecoherence have a severe disagreement. Via git blame, the [original implementation](https://github.com/SNEWS2/snewpy/commit/7cf6b41d62a220b5904296d0a8d90813ca7e33d4) is from you @sybenzvi, while @jpkneller [changed it](https://github.com/SNEWS2/snewpy/commit/73e81916382d1bb27662aef03e104a2d52826a1c) to the current value. Can the two of you agree on which one is the true TwoFlavorDecoherence?

sybenzvi:
> And in `test_2fd()`:
> 
> ```
> >       assert(xform.prob_ee(t, E) == 0.5)
> E       assert <Quantity 0.15956689> == 0.5
> ```
> 
> This looks like the test and the implementation of TwoFlavorDecoherence have a severe disagreement. Via git blame, the [original implementation](https://github.com/SNEWS2/snewpy/commit/7cf6b41d62a220b5904296d0a8d90813ca7e33d4) is from you @sybenzvi, while @jpkneller [changed it]

Something looks off here. The `TwoFlavorDecoherence` model has two mixing angles and seems to match the contents of the `AdiabaticMSW` transformation. Maybe it was a cut and paste error.

PR #44, opened 2021-07-15
Included some 3D models with SASI

mcolomermolla:
These models can be run with model class Bollig_2016 (same format)


JostMigenda:
Great to see even more models; thanks! 🤩

A couple of comments on the PR:
* [ ] It looks like the file `models/Tamborra/README.md~` got added by accident; could you delete that?
* [ ] You have included various minor changes to `from_snowglobes.py` and `to_snowglobes.py`, which look unrelated to this new model—probably also something that you added by accident. Could you please revert those?
* [ ] You note that this is the same file format as the Bollig_2016 models. Could you also add that to the README file?
* [ ] In all flux files I’ve checked, the final line is duplicated. Would probably be better to delete that (I think some interpolation algorithms do fail on duplicate values) and, perhaps, to let Irene Tamborra and/or the people running the Garching CCSN archive know so they can fix their hosted files.

And some (for lack of a better word) _structural_ comments, where I would love it if one of the main developers could weigh in:
* [ ] Since this is a collection of several different models, I think it might be better to put them into separate folders (`Tamborra_2014`, `Walk_2018`, `Walk_2019`)? Both for consistency with other models and to avoid confusion if Irene Tamborra ever publishes another model with a different file format.
* [ ] Also for consistency, I think it would be good to add entries for these models to [`model_class_dict`](https://github.com/SNEWS2/snewpy/blob/main/python/snewpy/to_snowglobes.py#L31). We could either map them directly to the `Bollig_2016` class or map them to classes named `Tamborra_2014` (and so on), and then make those an empty subclass of `Bollig_2016` in `models.py`. (Personally, I’d advocate for the latter: It may be one more line of code, but it is clearer and makes the same consistent interface available to any software that imports from snewpy.models—including sntools.)
* [ ] I noticed that this PR is targeted at the `release-version-1.0` branch, which seems to be rather stale and hasn’t seen any activity in the past few months. I guess it’s probably better to retarget it at the `main` branch?

mcolomermolla:
Many thanks for your quick review! I will take into account your comments. 
Concerning these points: 

- You have included various minor changes to from_snowglobes.py and to_snowglobes.py, which look unrelated to this new model—probably also something that you added by accident. Could you please revert those?
- I noticed that this PR is targeted at the release-version-1.0 branch, which seems to be rather stale and hasn’t seen any activity in the past few months. I guess it’s probably better to retarget it at the main branch?

This was done in purpose: I started from the previous pull request I made, which includes the possibility of getting directly the output files with the time series to be used by snewpdag. We have now some students which will include some analyses (as the fourier transform for SASI) into snewpdag, this is why I wanted the first implementation which provides the lightcurve as output and the new models together in the same branch for them to use it. 
Moreover, the released version is the one available publicly for use, while not the main. Also as you saw the tests are not passed in the main but were passed for the release. Finally, the main does not contain the SNEWPY.py script which allows to run all the chain together and makes is easy to be used. 
So, maybe I will keep this separate branch anyway, and push just the models to the main separately.

Concerning this:

- In all flux files I’ve checked, the final line is duplicated. Would probably be better to delete that (I think some interpolation algorithms do fail on duplicate values) and, perhaps, to let Irene Tamborra and/or the people running the Garching CCSN archive know so they can fix their hosted files.

Okey, I will check that. It might be something artificial happening in my script to convert the original Garching files to the format needed for snewpy.

evanoconnor:
Thanks Marta,

Regarding the models.  I assume you have permission to use these models, but do you have permission from the authors to reformat and post them publicly here?  If not, we can't include them.

from_snowglobes returns all of the data in a table (send `return_tables=True` as an option).  With this table you can completely reproduce all of the data in the files and therefore it should be good enough for what you need (see example scripts in doc/nb/).  If not, let us know what data you need and we can include it.  We want to keep to/from/run_snowglobes as generic as possible.  The changes you have here break that (for example, excessive print statements inside from_snowglobes)

as Jost said, that branch is stale and not the format of snewpy going forward.  The main branch can accomplish anything (and more) that that branch could, it just requires a slightly different setup.  The SNEWPY script can be very easily carried over to the new branch with very few changes, there are lots of other examples in the doc/nb/ folder too.

Yes, we need to work on the CI

mcolomermolla:
Many thanks for your inputs!

For the models, I do have permission to use them, but it is true I did not ask specifically for their post in snewpy, I could do that as we are planning to use them within SNEWS. 

Okey, then once I get the permission, I will make a new pull request into the main with the new models. 

Concerning the modifications to have a different ouput, it is true that maybe I forgot to remove some prints that I used for my own purpose before making the commit, but the change I want to make should still keep from_snowglobes as generic as possible, I just want to output in a file:
print( t, "\t", data, file=fout)
with data being the total number of events detected in this time bin for the smeared+weighted 
I implemented it with data being returned by the add_funct function in from_snowglobes, 
and the data being defined in add_funct as: result += sum(list(new_dict.values()))

This is all the changes needed, and it will avoid needing a second script to read the tables looping over energy bins and channels, but if this is not possible, we will adapt snewpdag to read the tables instead, which I know can be done too.
I can try to make a new proposal of this change in a new pull request to the main, and just close the old one with the prints and useless additions. Then you can see if this keeps from_snowglobes generic enough.

PR #46, opened 2021-07-23
remove constant factor from Warren_2020 model

JostMigenda:
> Note that the x neutrino luminosity is for **one** neutrino flavor - to get the total mu/tau neutrino and antineutrino luminosities requires multiplying this number by 4.

(from https://doi.org/10.5281/zenodo.3667908)

Accordingly, the nu_x luminosity should not be divided by 4 when reading in data from these files. This PR removes that factor and, while I’m at it, adds the journal reference and the DOI link for the neutrino fluxes.

### DIFF
--- a/python/snewpy/models.py
+++ b/python/snewpy/models.py
@@ -877,7 +877,8 @@ class Zha_2021(SupernovaModel):
 
 
 class Warren_2020(SupernovaModel):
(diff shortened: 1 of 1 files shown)


JostMigenda:
Merging now, as discussed in the meeting a few minutes ago.

PR #49, opened 2021-07-27
remove photospline import

JostMigenda:
Cleanup: the FornaxModel code, which used this import, was removed in #47


sybenzvi:
Agreed that we should dump photospline in favor of using the simple binned spectra in the Fornax model(s). We can add this dependency back in later if desired.

sybenzvi (approved):

PR #55, opened 2021-07-28
some fixes for units consistency in generate_time_series

mcolomermolla:


python/snewpy/to_snowglobes.py
```diff
@@ -40,16 +40,18 @@ def generate_time_series(model_path, model_file, model_type, transformation_type
...
-        dt = (tmax - tmin) / (ntbins+1)
+        dt = (tmax/u.s - tmin/u.s) / (ntbins+1)
     else:
         ntbins=30
-        dt = (tmax - tmin) / (ntbins+1)
+        dt = (tmax/u.s - tmin/u.s) / (ntbins+1)
 
-    tedges = np.arange(tmin, tmax, dt)
-    times = 0.5*(tedges[1:] + tedges[:-1])
+    tedges = np.arange(tmin.u.s, tmax/u.s, dt)
```
JostMigenda, line 53:
```suggestion
    tedges = np.arange(tmin/u.s, tmax/u.s, dt)
```

JostMigenda:
I just tried to test this change using `doc/nb/SNEWPY_examples.ipynb` with a slight change to use `generate_time_series`:
```
# tarredfile = to_snowglobes.generate_fluence(modeldir, filename, modeltype, transformation, distance, outfile,tstart,tend)
tarredfile = to_snowglobes.generate_time_series(modeldir, filename, modeltype, transformation, None, distance, outfile, None, None)
```

Unfortunately, while `to_snowglobes.generate_time_series` now completed, the next step (`run_snowglobes.go`) crashed.
After investigating, I noticed that `generate_time_series` produces malformed SNOwGLoBES flux files. E.g. the resulting `snowglobes/fluxes/s17.dat` file begins:
```
# TBinMid=1.46356 ssec TBinWidth=0.0579244s EBinWidth=0.2MeV Fluence at Earth for this timebin in neutrinos per cm^2
# E(GeV)        NuE     NuMu    NuTau   aNuE    aNuMu   aNuTau
   1.38589050E-13   1.26777533E-04 1 / s   6.62679561E-05 1 / s   6.62679561E-05 1 / s   4.13778389E-05 1 / s   1.08967803E-04 1 / s   1.08967803E-04 1 / s
   2.00000000E-04   4.76669199E+05 1 / s   2.50438463E+05 1 / s   2.50438463E+05 1 / s   1.55745132E+05 1 / s   4.09699690E+05 1 / s   4.09699690E+05 1 / s
   4.00000000E-04   9.64952907E+05 1 / s   5.14187780E+05 1 / s   5.14187780E+05 1 / s   3.16718358E+05 1 / s   8.29455377E+05 1 / s   8.29455377E+05 1 / s
```
… which then causes SNOwGLoBES to crash when it tries to parse the file to calculate event numbers.

I think the root problem is that after these changes, `dt` is dimensionless instead of having the unit seconds. [When multiplying `osc_spectra` by `dt`](https://github.com/SNEWS2/snewpy/blob/641cf82d4fd8e1f1fb6e77c04c62c7d3df8aa7c1/python/snewpy/to_snowglobes.py#L89), the result then has the unit 1/s, which gets written to the file. Does that make sense or am I missing something?

If the problem is simply that `np.arange()` can’t deal with `astropy.Quantities` as arguments, maybe the better fix would be to leave `dt` as is and just change that one line to `tedges = np.arange(tmin/u.s, tmax/u.s, dt/u.s)`?
(In my testing that appears to work, but then I run into another crash during `collate()`; but I think that’s unrelated and I’ll note it in #88 instead.)

evanoconnor:
I'm running across this now.  the problem with `tedges = np.arange(tmin/u.s, tmax/u.s, dt/u.s)` is that the next line, which is the times, is not dimensional now.  I suggest:

`tedges = np.arange(tmin/u.s, tmax/u.s, dt/u.s)*u.s`

I'm working through the collate error now, I will test and work through the time units there

mcolomermolla:
Actually, times does not have to be adimensional, otherwise you run into an error when calling flavor_transformation in generate_time_series. At least I found this error. I did not get any problem with generate_time_series after these fixes, but it is true that I did not run snowglobes after it. Thanks for having checked that!

JostMigenda:
Looks like this is fixed in #87, so I’ll close this PR here. Thanks for your help on this!

PR #56, opened 2021-07-28
Include new 3D models with SASI

mcolomermolla:


JostMigenda:
Very nice to add new models, thanks a lot! 🤩 Just two quick comments:
* It looks like the issue with the last line being duplicated in each file is still there. Could you fix that?
* Could you revert the commit dealing with units in generate_time_series? I think it’s better to discuss that separately in #55.

PR #60, opened 2021-08-01
Sybenzvi/fornax2dnormfix

sybenzvi:


sybenzvi:
This should fix the normalization issue described in #59.

Issue #51, opened 2021-07-27
make SNEWPy available on PyPI


JostMigenda:
[Moved this from #45 into a separate issue.]

Making SNEWPY available on PyPI so people can install it with a single `pip install snewpy` is good for usability and would make it a lot easier for me to integrate sntools with SNEWPY. (PyPI doesn’t let packages like sntools install dependencies from a GitHub repo, so users currently have to install SNEWPY separately after installing sntools.)

Packaging and uploading to PyPI can be largely automated with GitHub Actions; I’ve done that for sntools and could contribute it here, too.

Another point that we should think about is whether to include the SN models? I’d tend towards not including them, so people don’t have to download all 100s of MBs of model files on every install; even if they just want use a single model. Instead, maybe it’s worth adding a "setup" function that lets users select which models to download after install?

JostMigenda:
Available at https://pypi.org/project/snewpy/ now! 😊

PR #66, opened 2021-08-04
Lazy downloads of model files & PyPI workflow

### LINKED ISSUE #53: Enable lazy downloads of model datasets.
Some of the model data sets uploaded to GitHub are fairly large (>20 MB) and should not be part of a core snewpy package downloaded from pypi (see issue #51). We may want to make a model registry and only download the data when the user requests the model. This behavior is enabled in the [sncosmo package](https://sncosmo.readthedocs.io/en/stable/), for example.

JostMigenda:
This closes #53 by adding two functions:
* `_get_model_urls()`: This creates a dictionary of available models in the current release and the URLs at which these files are available. (It assumes that the Git tag of a release is simply `v` plus the version number in `_version.py`, e.g. `v1.0.0`.)
* `get_models()`: After installing snewpy from PyPI, a user should run this function to download model files. It presents a straightforward interface to select which models to download; alternatively, you can use an optional argument for that (e.g. `get_models(models=['Bollig_2016', 'Warren_2020'])`

This PR also adds a workflow which demonstrates how to use these functions and which could immediately upload the build product to PyPI (laying the groundwork for #51, though those lines are currently commented out). See the [workflow outputs](https://github.com/SNEWS2/snewpy/runs/3245468399), particularly the `build` step.

Before merging this PR, three minor things still need to be cleaned up:
* [x] For testing purposes, I’ve hard coded an existing version tag in `_get_model_urls()`
* [x] ~There’s some debug code in `publish.yml`~ [EDIT: This is a good check to confirm that the build is working; so let’s leave it as is.]
* [ ] We should add a note on using `get_models()` to the README. However, we should only add this when we also change installation instructions to get snewpy from PyPI; right now, when people are supposed to clone the repo first (including all model files), this lazy download doesn’t make sense and would be confusing.


JostMigenda:
@sybenzvi From what I can tell based on their documentation, my approach is a bit different from the sncosmo example you gave. (Though I’d argue that both downloading models ahead of time and referring to them based on file names (which can be autocompleted, etc.) are preferable to what sncosmo is doing.)
If you have comments or suggestions for improvement, let me know!

JostMigenda:
With today’s changes, this PR should also close #51 (apart from documentation and setting up the token for PyPI). Brief summary:

* Now pushes to TestPyPI (see https://test.pypi.org/project/snewpy/)
* Bugfix for models with nested directories (h/t @mcolomermolla)
* Bugfix for displaying the README file on PyPI correctly.
* Minor change to ensure that the workflow raises an error if the download URLs are invalid. That means it won’t accidentally publish a broken version to PyPI, e.g. if the `__version__` in the code doesn’t match an existing tag. (Demo: [mismatch](https://github.com/SNEWS2/snewpy/actions/runs/1102457597), [matching versions](https://github.com/SNEWS2/snewpy/actions/runs/1102473200))
image

* Since @mcolomermolla noted that downloading all models is slower than cloning the repo, I’m using multithreading to download files now. This should be significantly faster, particularly for models with a large number of small files, like Warren_2020.

JostMigenda:
As discussed at the meeting a few min ago, I’ll merge this PR this afternoon and tag v1.1b1 immediately afterwards to push that to PyPI.

PR #67, opened 2021-08-05
snewpy to snewpdag

mcolomermolla:
Script to run full snewpy chain and produce input file for snewpdag


python/snewpy/scripts/snewpy_to_snewpdag.py
```diff
@@ -0,0 +1,67 @@
...
+from argparse import ArgumentParser
+from snewpy import to_snowglobes
+from snewpy import run_snowglobes
+from snewpy import from_snowglobes
+import tarfile
+import numpy as np
+import os
+from astropy.io import ascii
+from astropy import units as u
+import matplotlib.pyplot as plt
```
JostMigenda, line 11:
```suggestion
from snewpy import to_snowglobes, run_snowglobes, from_snowglobes
import numpy as np
from astropy import units as u
```
removing unused imports

python/snewpy/scripts/snewpy_to_snewpdag.py
```diff
@@ -0,0 +1,67 @@
...
+#now run snowglobes, this will loop over all the fluence files in `tarredfile`
+run_snowglobes.go(SNOwGLoBES_path, modeldir, tarredfile, detector_input=detector,verbose=False)
+print("Done snowglobes...")
+
+#now collate results of output of snowglobes
+print("Collating...")
+tables = from_snowglobes.collate(SNOwGLoBES_path, modeldir, tarredfile, detector_input=detector,skip_plots=True,return_tables=True,verbose=False)
+
+#read results from snowglobes and put lightcurve in output file for snewpdag
+fout = open("/path/to/output/snewpy_output_"+detector+"_"+modeltype+"_"+filename+"_1msbin.txt", "a")
```
JostMigenda, line 61: It may be nicer to move `/path/to/output/` into a separate variable that’s defined at the top of the file, right after `SNOwGLoBES_path` and `SNEWPY_models_base`; that way, users will find all the paths they need to adjust in one place.

mcolomermolla: yes, agree!

python/snewpy/scripts/snewpy_to_snewpdag.py
```diff
@@ -0,0 +1,61 @@
+
+from snewpy import to_snowglobes, run_snowglobes, from_snowglobes
+import numpy as np
+from astropy import units as u
+
+SNOwGLoBES_path = "/location/of/snowglobes/" #where snowglobes is located
+SNEWPY_models_base = "/location/of/models/" #where models (aka input for to_snowglobes) is located
+output_ptah = "/path/to/output/" #where the output files will be located
```
JostMigenda, line 8:
```suggestion
output_path = "/path/to/output/" #where the output files will be located
```
typo

mcolomermolla:
I will update this pull request making a new one from the current main. 
When this request was done, there was no snowglobes module with its corresponding functions.

PR #68, opened 2021-08-05
set long_description_content_type for README.md

JostMigenda:
Will render poorly on PyPI otherwise; see https://test.pypi.org/project/snewpy/0.1.0.dev2/ and https://packaging.python.org/specifications/core-metadata/#description-content-type-optional


JostMigenda:
Sorry; meant to do that as part of #66 to test it there first.

PR #77, opened 2021-08-11
Sybenzvi/updatewalkmodels

sybenzvi:
Added more class-specific details and demo notebooks for the `GarchingArchiveModel` child classes based on simulations from Walk et al., 2018 and 2019.


JostMigenda (approved):
LGTM

PR #80, opened 2021-08-12
Sybenzvi/more test cleanup

sybenzvi:
Additional fixes to unit tests.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,6 +1,6 @@
 # This is a basic workflow to help you get started with Actions
 
-name: SNEWPy Test Suite
+name: tests
 
 # Controls when the action will run. Triggers the workflow on push or pull request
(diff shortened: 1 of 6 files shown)


JostMigenda:
Thanks for these changes!
Since you moved all the tests to unittest in #79, would it make sense to update the test workflow to use unittest, too? Or should we stick with Pytest for running them?

sybenzvi:
> Since you moved all the tests to unittest in #79, would it make sense to update the test workflow to use unittest, too? Or should we stick with Pytest for running them?

Good point. I think we probably don't need pytest anymore.

sybenzvi:
> Good point. I think we probably don't need pytest anymore.

On second thought @JostMigenda , I'll leave it alone. I don't think it harms anything. If you want to change it you can. Otherwise I say we merge.

Issue #76, opened 2021-08-11
Write documentation


JostMigenda:
Segev and I have set up [documentation for SNEWPY](https://snewpy.readthedocs.io/en/latest/index.html).
In #72 I’ve added a brief "Getting Started" page and filled in the rest with autogenerated content from docstrings. Now we need to fill in the missing parts! In particular:

* [x] On the Getting Started page, we need a few brief Usage examples that don’t need SNOwGLoBES. (We can probably take something from one of the Jupyter notebooks here?)
* [x] The "Using SNEWPy as a Front End for SNOwGLoBES" page is almost empty. It needs:
    * [x] Information on the required SNOwGLoBES version
    * [x] A description of the basic workflow
    * [x] Docstrings for `to_snowglobes`, `run_snowglobes` and `from_snowglobes`
* [x] Edits to the Supernova Models page, e.g. as listed in the "Warnings" box at the top.
* [x] Edits to the Flavor Transformations page, e.g. as listed in the "Warnings" box at the top.
* [x] Probably some additions/edits to the `snewpy.neutrino` page as well?
* [x] Add a Community Guidelines section [as requested by JOSS](https://joss.readthedocs.io/en/latest/review_criteria.html#documentation). Brief info on how to contribute, report issues, seek support.
* [x] A README file in `doc/nb/` with brief descriptions of the various notebooks.

I’ll keep working on this occasionally; but need help, particularly from people who have more experience using SNOwGLoBES!

JostMigenda:
* [ ] Add a Community Guidelines section [as requested by JOSS](https://joss.readthedocs.io/en/latest/review_criteria.html#documentation). Brief info on how to contribute, report issues, seek support.

(I can write that; just making a note here so I don’t forget.)

JostMigenda:
README file in `doc/nb/` added in #117.

PR #87, opened 2021-08-19
merge to/run/from_snowglobes into a single module

### LINKED ISSUE #82: Cleanup of {to,run,from}_snowglobes
(This is not a big deal in the grand scheme of things, but we should make these changes before we publish the ApJ paper; afterwards, changing the API to be different from what’s described in the paper will lead to user confusion.)

* The main function of `run_snowglobes` should be called `run` instead of `go`. (It seems confusing to me to use different verbs in the function name and module name; and of the two, _run_ is preferable because you don’t _go_ SNOwGLoBES, you _run_ SNOwGLoBES.
* Having separate arguments for the folder containing the tarfile and the tarfile itself is unnecessary. This should just be one argument; if the components are ever needed separately, `os.path.split` can easily do that.
* The go/run function needs a brief docstring, to help with #76.

I’ll try to work on this early next week; just wanted to write it down already so we don’t forget.

JostMigenda:
This merges `to_snowglobes`, `run_snowglobes` and `from_snowglobes` into a single module (`snewpy.snowglobes`). I’m leaving stubs in the original files to ensure existing code doesn’t break. Additional renaming and clean-up will come in a bit; once ready, this will fix #82.

Throughout, I will update `doc/nb/SNEWPY_examples.ipynb` to demonstrate the changes; and to verify that the output is unchanged.


JostMigenda:
These commits
* add docstrings and some basic documentation (which helps with #76)
* simplify the API by having a single argument for the file path
* rename `go` to `simulate`.

I think it’s ready for review now, but I will add another commit later that updates all the demo notebooks.

evanoconnor:
Hi, sorry for my silence.  I'm okay with this cleanup.

evanoconnor:
A note about `snowglobes.generate_time_series()`.  It is quite different from `snowglobes.generate_fluence()` and they are not really swappable, they do different things.  We will need to clearly describe what each one is doing carefully.

evanoconnor:
This can replace #55, but in the new snowglobes module format
This should resolve #90 as well

evanoconnor:
This fixes the `SNEWS2.0_rate_table.py` script which was broken (#88)

JostMigenda:
Thanks for the updates @evanoconnor!

My first commit cleans up some files that looked like they were added accidentally.
My second commit works with any input file extension, so we don’t have to remember to update the code if we e.g. add a new model with `.csv` (or `.txt` or …) input files. (The same for `generate_fluence`, while I was working on it.)

I’ll update `SNEWS2.0_rate_table.py` next, now that it should work.

JostMigenda:
Updated `SNEWS2.0_rate_table.py`. I removed the KM3NeT entry for now, since sample scripts should work with the main branch of SNOwGLoBES. Once [the relevant PR](https://github.com/SNOwGLoBES/snowglobes/pull/10) is merged into SNOwGLoBES, we can re-add this.

JostMigenda:
Modernized `TimeSeries.py`. In addition to merging the SNOwGLoBES-related modules, I’ve also verified that it works thanks to Evan’s earlier changes; so this fixes #88 as well.

It might still be better to merge it into `doc/nb/SNEWPY_examples.ipynb` instead and add explanations about the difference between `generate_fluence` and `generate_time_series`, but that’s beyond the scope of this PR.

PR #92, opened 2021-09-07
rename `get_oscillatedspectra` to `get_transformed_spectra`

JostMigenda:
The function name `get_oscillatedspectra` is inaccurate, since it covers a wide variety of `FlavorTransformation`s—including non-oscillatory ones like neutrino decay. This PR renames it to `get_transformed_spectra`. (The ApJ(S) paper draft was already updated to the new name.)

All existing uses of the function in the code and included example notebooks are updated. For users who have custom scripts, I’ve left a stub under the old function name which prints out a warning message and redirects to the new function name.


mcolomermolla:
If you rename get_oscillatedspectra, maybe we should do the same with get_initialspectra to be consistent?

JostMigenda:
So, change `get_initialspectra` to `get_initial_spectra`? That would definitely be nicer, yes. I just wasn’t sure whether this minor change is worth the effort; but I can do it, too.

JostMigenda:
Renamed `get_initialspectra()` to `get_initial_spectra()` as suggested. As above, all existing uses of the function in the code & example notebooks are updated and I’ve left a stub under the old function name.

PR #93, opened 2021-09-07
Update flavor_transformation.py

jpkneller:
Added some comments to the four-flavor prescriptions.


jpkneller:
You're right. Thanks for catching that.

PR #98, opened 2021-09-10
better handling of model names and corresponding subclasses

JostMigenda:
This PR renames the Fornax `SupernovaModel` subclasses so that the `model_type` parameter (in the `generate_*` functions in `snewpy.snowglobes`), subclass name and name of the directory with model flux files are all consistent.

It also introduces an easier way to link `model_type` to the subclass with that name. When adding a new `SupernovaModel` subclass, it is no longer necessary to add it manually in both `generate_*` functions. This saves us a little bit of manual labour and will also make bugs like #95 impossible in the future.


JostMigenda:
Once this is merged, we need to remember to update the class names in the paper draft accordingly.

PR #105, opened 2021-09-13
Vectorize NeutrinoDecay

### LINKED ISSUE #31: Vectorize NeutrinoDecay calculation in flavor transformation module
The NeutrinoDecay transformation is extremely slow; the loop over energy spectra can be vectorized.

sgriswol:
These changes compute the NeutrinoDecay probabilities using numpy broadcasting. It should run considerably faster. Running Flavor_XForm.ipynb locally, the cells associated with NeutrinoDecay ran 86 times faster (evaluated by ipython `timeit` magic command).

Fixes #31, closes #34

### DIFF
--- a/python/snewpy/flavor_transformation.py
+++ b/python/snewpy/flavor_transformation.py
@@ -1144,23 +1144,14 @@ class NeutrinoDecay(FlavorTransformation):
         prob : float or ndarray
             Transition probability.
         """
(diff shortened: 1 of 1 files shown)


sybenzvi (approved):
Verified that the changes do not affect the output of the Flavor transformation.

PR #108, opened 2021-10-05
Add a new BH simulation with Togashi EOS

nuberoi:
Create a new FITS file for a recently-added Nakazato simulation - Togashi EOS for a BH-forming 30M_sun progenitor. The data were obtained from the [Supernova Neutrino Database](http://asphwww.ph.noda.tus.ac.jp/snn/) ([Nakazato et. al.](https://arxiv.org/abs/2103.14386))


JostMigenda:
This looks great, thanks a lot! I did verify (using Nakazato_2013.ipynb) that we’re able to reproduce the luminosity plots from https://arxiv.org/abs/2103.14386 and have just updated the README file to include the new paper.

PR #112, opened 2021-10-13
Update README and documentation

### LINKED ISSUE #111: pip installation from PyPI leads to failed tests
Hello SNEWPY team! This is part of the JOSS review at https://github.com/openjournals/joss-reviews/issues/3772.

Using the suggested installation from the README (`pip install snewpy` and then `python -c 'import snewpy; snewpy.get_models()'` to grab models), the tests fail. Specifically, `test_models.py` outputs

```
FileNotFoundError: [Errno 2] No such file or directory: 'models/Nakazato_2013/nakazato-shen-z0.004-t_rev100ms-s13.0.fits'
```

I think this is just because when installing as directed in the README, the package distributed by PyPI doesn't have all of the included files in the github repo, and then when downloading the files the models go to a directory called `SNEWPY_models` instead of where the test script is looking (`models/`).

I think this is all okay, because I understand wanting to have only a limited version of the code available on pypi so that you don't need all of the large files from the examples / documentation / models. However, there is no alternative method described in the docs on how to install the code that would allow the user to run examples and run tests. Could you include a brief set of installation instructions in the README for those that would want access to the examples and such? I'm thinking something along the lines of what you do in the GitHub actions to automate tests, eg
```
git clone {SNEWPY_git_url}
python setup.py install --user
```

JostMigenda:
This should fix #111.

It also includes some improvements to the JOSS paper in response to @apizzuto’s feedback in openjournals/joss-reviews#3772.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ This is a walkthrough to install GLoBES and SNOwGLoBES locally in the users home
 ## Usage and Documentation
 Example scripts which show how SNEWPY can be used are available in the
 `python/snewpy/scripts/` subfolder as well as notebooks in `doc/nb/`.
+Most downloadable models also include a Jupyter notebook with simple usage examples.
 
 A paper describing SNEWPY and the underlying physics is available at [arXiv:2109.08188](https://arxiv.org/abs/2109.08188).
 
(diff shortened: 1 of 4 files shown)


sybenzvi (approved):
Doc updates look good. Comment: in addition to `doc/nb` a lot of the model notebooks have been moved into the `models` subfolder, so we should indicate that.

PR #123, opened 2021-11-01
fix reading from SNOwGLoBES files

### LINKED ISSUE #122: Digits removed when collating files
Dear All,

I have installed the following commit of snewpy: 7ac532a4f4cd6723e53e3f41ed023ed542f63373 and modified python/snewpy/scripts/TimeSeries.py to keep all the generated files, using:

snowglobes.collate(SNOwGLoBES_path, outfile, detector_input="icecube", skip_plots=True, remove_generated_files = False)

When I compare the collated files to the original files, I find that, for a few rate values and for some files, the first digit of the result has been removed. See for example the IBD rate for E = 0.002739 GeV in:

nakazato-LS220-BH-z0.004-s30.0.tbin9.AdiabaticMSW_NMO.-0.110,0.342,30-10.0kpc_ibd_icecube_events_smeared_unweighted.dat  (rate = 3.65327e-05)

compared to

Collated_nakazato-LS220-BH-z0.004-s30.0.tbin9.AdiabaticMSW_NMO.-0.110,0.342,30-10.0kpc_icecube_events_smeared_unweighted.dat (rate = 6.5327e-06)

Best regards,
Sonia

JostMigenda:
Fixes #122, where SNEWPY would cut off the initial digit of the event count when reading SNOwGLoBES files.

@soso128: In my testing, this has been fixed. Can you please check out this branch and confirm whether it fixes the issue?

### DIFF
--- a/python/snewpy/snowglobes.py
+++ b/python/snewpy/snowglobes.py
@@ -866,51 +866,28 @@ def collate(SNOwGLoBESdir, tarball_path, detector_input="all", skip_plots=False,
                 name_iteration = "{0}_{1}*{2}*{3}*".format(flux, input_val, detector, smear)
                 if fnmatch.fnmatch(afile, name_iteration):  # and "Collated" not in str(afile):
                     FilesToCleanup.append(homebase+"/"+afile)
-                    fileinsides = open("{0}/{1}".format(homebase, afile))  # fine
-
-                    lines = fileinsides.readlines()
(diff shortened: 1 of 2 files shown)


JostMigenda:
Of course, ignoring the initial digit reduces the number of events; so this fix increases the number of events in the integration test slightly (from 4044.84127 to 4044.84174), which breaks the test. This second commit should fix that.

PR #127, opened 2021-11-04
WIP: Sheshuk/refactoring

### LINKED ISSUE #126: Snowglobes interface refactoring
As I understand snowglobes interface `python/snewpy/snowglobes.py` inherits from several scripts. 
It works could use refactoring. It contains:
* Repetitive lines, which can be substituted by a cycle
* Multiple checks `if verbose: print...` it's better to use `logging` module
* Opening and closing files instead of using context manager `with open(filename) as f`

Sheshuk:
Closes #126


Sheshuk:
Abandoned, using another branch: #132

Issue #120, opened 2021-10-27
Add value checks to Warren_2020 model (and others?)


sybenzvi:
From @jpkneller:

> A grad student here at NCSU found a negative mean energy in the M=55 a=1.25 Warren_2020 model. It's for the nu-x flavor at t = 0.00959799. The way the Warren_2020 model is written it takes the log of this mean energy, which causes the code to choke. I'm wondering if we should put in a few if statements to return zero if any mean energy or luminosity is <= 0.

Leaving this as an uncaught divide by zero exception or `ValueError` is not great. I think we could return zero, but we should emit warnings too so that the user isn't quietly getting unexpected output, thinking everything is fine and dandy.

PR #129, opened 2021-11-05
Fix div by zero in Warren_2020 models.

sybenzvi:
Add a check for values of average energy, pinch, and luminosity that can produce NaN in spectra.


sybenzvi:
This PR will address issue #120.

python/snewpy/models.py
```diff
@@ -1211,6 +1211,11 @@ def get_initial_spectra(self, t, E, flavors=Flavor):
             Ea = get_value(np.interp(t, self.time, self.meanE[flavor].to('erg')))
             a  = np.interp(t, self.time, self.pinch[flavor])
 
+            # Sanity check to avoid invalid values of Ea, alpha, and L.
+            initialspectra[flavor] = np.zeros_like(E, dtype=float) / (u.erg*u.s)
+            if L <= 0. or Ea <= 0. or a <= -2.:
+                return initialspectra
```
JostMigenda, line 1217: This code is inside the `for flavor in flavors` loop, so returning `initialspectra` here could mean we only return values for some flavors, which might crash the calling code, if it expects otherwise.

I think something like
```suggestion
            if L <= 0. or Ea <= 0. or a <= -2.:
                initialspectra[flavor] = np.zeros_like(E, dtype=float) / (u.erg*u.s)
                continue
```
would be better?

JostMigenda (changes requested):

python/snewpy/models.py
sybenzvi: Quite right, that was a boneheaded mistake. Just committed a fix.

JostMigenda (approved):

PR #135, opened 2021-11-11
Adding png image to be used for html documentation

### LINKED ISSUE #134: Image `luminosity-comparison.pdf` in documentation: not found
Looks like the plot in our documentation is not displayed:
https://snewpy.readthedocs.io/en/latest/gettingstarted.html#usage

I get the same when building the docs with `make html`: the file is there, but it's not displayed.
It works with `make latexpdf` though. So, probably it's a file type problem. We should generate a `png` plot as well.

Sheshuk:
Fixes #134


JostMigenda:
Oh, right … there are browsers that don’t display pdf figures inline. 🙄

Thanks for discovering & fixing this!

JostMigenda (approved):

PR #138, opened 2021-11-12
Adding preSN models

### LINKED ISSUE #121: Interface preSN models into snewpy
I would like to add preSN models (https://github.com/SNEWS2/presn-models) to the snewpy models.  I would also like to run time series through SNOwGLoBES for varying start times before collapse.

I am willing to do the work, but don't know the best format for the models' spectra and times.  Is there anyone that would like to help me add these features to snewpy?

Sheshuk:
Closes #121

Changes moved to #159 
----------------------
* Reorganized the models module: splitting to `models.ccsn`,
* Moved base classes to `models.base`, added base class `PinchedModel`
* Removed the repeating methods `get_initial_spectra` to the base class where possible
* Moved printing methods `__str__` and `_repr_markdown_` to the base `SupernovaModel` class. Now they print the table from the `self.metadata` dict, so one can fill it in the subclass model's constructor. 
* Added deprecation warning on trying to import ccsn models directly from `snewpy.models`
 
Changes in this PR in addition to #159
---------------------------------------

* Added three presupernova models to `models.presn`
* Added models data as a git submodule (for now)


JostMigenda:
Just a note that we commonly use things like `from snewpy.models import Bollig_2016` in sample code (e.g. the various Jupyter notebooks included in the repo) and have also described this structure in the ApJ paper. Making them only accessible under `snewpy.models.ccsn` would thus be a breaking change.

I’d suggest adding code in `models/__init__.py` so that the models from `models/ccsn.py` also remain available under their old name. (At least for now: We can add a deprecation message when importing a model directly from `snewpy.models`—see [the first code example in PEP 562](https://www.python.org/dev/peps/pep-0562/#rationale)—and then clean that up in the next major version of SNEWPY.)

Sheshuk:
@JostMigenda I've implemented your suggestion, but using `UserWarning` class - for some reason I don't get warning messages when I do `DeprecationWarning`. Maybe you can help to make it work?
For the rest I think the PR is ready. I didn't make unit tests for new models yet, since we're not decided on the models bookkeeping.

And making preSN models work with `generate_fluence/generate_time_series` is for #152

JostMigenda:
Ah, yes. `DeprecationWarning`s are hidden by default in most cases, see [PEP-565](https://www.python.org/dev/peps/pep-0565/) for background. From the `warnings` [module documentation](https://docs.python.org/3/library/warnings.html?highlight=deprecated#warning-categories), it looks like `FutureWarning` would be most appropriate for our case; but the default `UserWarning` would be fine, too.

python/snewpy/models/presn.py
```diff
@@ -0,0 +1,141 @@
...
+
+def _interp_E(e0, v0, axis=1):
+    "linear interpolation"
+    return interp1d(e0, v0, axis=axis, fill_value=0, copy=False, bounds_error=False)
+
+
+def _interp_TE(times, energies, array, ax_t=1, ax_e=2):
+    def _f(t, E):
+        a_t = _interp_T(times, array, axis=1)(t)
+        a_te = _interp_E(energies, a_t, axis=2)(E)
```
JostMigenda, line 25: Should this be `axis=ax_t` and `axis=ax_e` instead of the hard-coded values?

Sheshuk: Good catch! Thanks! Fixed in ac90339

mcolomermolla (approved):
No comments from my side on this commit. 
Agree with the changes and removal of the repetition of the get_time, get_... functions for every model, moved now to pinched model

mcolomermolla (approved):
Those changes are fine to me

mcolomermolla (approved):
Approved changes

mcolomermolla (approved):
Changes approved: 
- presn-models submodule works 
-  presn.py first version, modified in next commits

mcolomermolla (approved):
Those changes look fine to me. Just check why test is not passed

mcolomermolla (approved):

mcolomermolla (commented):
- get_time and get_initial spectra repeated for every model -> can't these two functions be put outside the models just once as you did for models.ccsn >

- here the tests are passed with the three models, so I would no worry that Patton implementation was not passing test in previous commit

Sheshuk:
@mcolomermolla Thanks for reviewing! I think your comments here ended up not connected to the exact commits. You can comment directly on the code lines in "Files changes" section, clicking on the plus sign near the code line.

mcolomermolla (approved):

mcolomermolla (commented):
Later changed to UserWarning and finaly by FutureWarning

mcolomermolla (approved):

mcolomermolla:
> @mcolomermolla Thanks for reviewing! I think your comments here ended up not connected to the exact commits. You can comment directly on the code lines in "Files changes" section, clicking on the plus sign near the code line.

oh okey, that's bad, I thought they would be associated to each commit, speacilly when approving changes...

mcolomermolla:
As previous comments were not properly associated to the commit, here the summary.
Basically, all the changes look good to me, just:

- in presn.py : get_time and get_initial spectra repeated for every model -> can't these two functions be put outside the models just once as you did for models.ccsn

- check for the failed test of the init file commit, the other failure is fine as it was corrected in latest commit with the presn models implemented.

For the rest, everything looks OK to me to accept the PR. I tested the backwards compatibility of my scripts after moving to models.ccsn and the presn submodules and everything works fine.

I will later check that preSN models work with generate_fluence/generate_time_series with implementation in #152

Sheshuk:
Thank you for the thorough review!
I think github applies "approve changes" to the pull request in total, so you don't have to approve each commit separately, but just check the final changes in files. As the individual commits show the intermediate changes, it's possible that some of them fail the tests - what's important is that in the latest commit (ac90339 in this case) all the tests are good.

> in presn.py : get_time and get_initial spectra repeated for every model -> can't these two functions be put outside the models just once as you did for models.ccsn

Yes, we're moving `get_time` to the base class SupernovaModel in #159, so I'll do it also in models.presn, after #159 is merged.
But as for `get_initial_spectra` - the implementation is different for each model, so we have to keep these functions.

PR #144, opened 2021-11-16
SNOwGLoBES interface cleanup

JostMigenda:
This PR
* cleans up some unused code and imports that were left over from re-writing the `collate` function
* moves all imports to the top of the module
* replaces the hardcoded name of the cache file (that is created in `simulate` and read in `collate`) with one that is analogous to the tarball file name


Sheshuk (approved):
Looking good! Thanks for this cleanup

PR #148, opened 2021-11-17
Model download improvements

### LINKED ISSUE #145: Problem downloading SN models: Warren a1.25 m120
When executing "python -c 'import snewpy; snewpy.get_models()'" many SN models download, but one fails:

Failed to download https://github.com/SNEWS2/snewpy/raw/v1.1b4/models/Warren_2020/stir_a1.25/stir_multimessenger_a1.25_m120.h5 to 'SNEWPY_models/Warren_2020/stir_a1.25/stir_multimessenger_a1.25_m120.h5'.

at the end it says:

ERROR: 1 exceptions occured. ([<HTTPError 502: 'Bad Gateway'>])

When I rerun it, it works fine and the 120M Warren model file is downloaded. Maybe it is hitting the server too fast? This isn't that important of an issue. Maybe you could add a comment if such this 502 error pops up that rerunning might help. Also, some files weren't attempted to download on the first pass and were only downloaded on the second pass.

JostMigenda:
This closes #145 by adding an improved error message if a model file download fails.

Additionally, now that the Jupyter notebooks demonstrating each model are downloaded together with the model files, this PR ensures that the notebooks are re-downloaded even when they already exist locally. That’s because unlike existing model files (which shouldn’t change to avoid confusion) these notebooks may be updated over time as more progenitors for that model are added or as the SNEWPY API evolves.

### DIFF
--- a/python/snewpy/__init__.py
+++ b/python/snewpy/__init__.py
@@ -95,6 +95,7 @@ def get_models(models=None, download_dir="SNEWPY_models"):
             exceptions.append(result.exception())
     if exceptions:
         print(f"ERROR: {len(exceptions)} exceptions occured. ({exceptions})")
+        print("Please check your internet connection and try again later. If this persists, please report it at https://github.com/SNEWS2/snewpy/issues")
         exit(1)
     pool.shutdown(wait=False)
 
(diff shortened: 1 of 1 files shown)


sybenzvi (approved):
Seems like a reasonable change.

Off-topic for this PR, seeing print statements makes me think we should switch to only using the logging module. I'll probably make an issue about it.

PR #153, opened 2021-11-22
Enable caching for pip, GLoBES and SNOwGLoBES

### LINKED ISSUE #137: Caching the dependencies for CI jobs
We are installing GLOBES and SNOWGLOBES for each integration test job. 
It's possible to use [caching action](https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows) to keep the directories. 
Another solution is to prepare a Docker container and use it to run the jobs, using [jobs.<job-id>.container](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idcontainer) option

Sheshuk:
Closes #137

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -51,7 +51,7 @@ jobs:
         id: snowglobes_tag
         run: |
           tag=$(git ls-remote --sort='version:refname' --tags --refs https://github.com/SNOwGLoBES/snowglobes.git | tail -1 | cut -d'/' -f3)
-          echo '::set-output name=latest::${tag}'
+          echo "::set-output name=latest::${tag}"
 
       - name: Caching SNOwGLoBES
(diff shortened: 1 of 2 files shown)


Sheshuk:
Hm, caching doesn't seem to work. Also looks like the most time is spent on downloading our models

Sheshuk:
As far as I understood, it should be possible to cache any directory. 
But it didn't work for me - I tried to get the GLoBES & SNOwGLoBES to cache and restore from the cached directory, but it keeps creating a new cache every time. Also skipping the `install dependencies` step in case of cache hit didn't work for some reason.

Overall, I think we should try out the Docker container variant. That way we can set all the needed dependencies there, and also optionally have all the models preinstalled there. I'm not sure what will be faster though - to download the models from github or to download entire container with models.

PR #156, opened 2021-11-25
asyncio fixes under Python 3.7

### LINKED ISSUE #155: Integration test fails under Python 3.7
As noted in #154, the integration test currently [fails under Python 3.7](https://github.com/SNEWS2/snewpy/runs/4327623529?check_suite_focus=true). The error is in `sng.run()`, which has led me to chase down an asyncio rabbit hole …

The initial problem was that `asyncio.create_subprocess_exec()` doesn’t accept `pathlib.Path` arguments. This appears to have [changed in Python 3.8](https://bugs.python.org/issue35246), though [the documentation isn’t updated](https://docs.python.org/3.11/library/asyncio-eventloop.html#asyncio.loop.subprocess_exec) so maybe we shouldn’t rely on that. Luckily, that one’s an easy fix; draft PR coming soon.

Unfortunately, once that’s resolved I run into another error: `sng.run()` returns `RuntimeError('Cannot add child handler, the child watcher does not have a loop attached')`
@Sheshuk, you appear to understand the guts of asyncio much better than I do—can you take a look at that?

JostMigenda:
This PR is intended to fix #155.

As discussed there, it only fixes the first part currently, so I’ll keep it as a draft.

### DIFF
--- a/python/snewpy/snowglobes_interface.py
+++ b/python/snewpy/snowglobes_interface.py
@@ -260,7 +260,7 @@ class Runner:
             with open(self.base_dir/'supernova.glb','w') as f:
                 f.write(cfg)
             #run the snowglobes process:
-            p = await asyncio.create_subprocess_exec('bin/supernova', self.flux_file.stem, chan_file, self.detector,
+            p = await asyncio.create_subprocess_exec('bin/supernova', self.flux_file.stem, str(chan_file), self.detector,
                 stdout=asyncio.subprocess.PIPE,
                 stderr=asyncio.subprocess.PIPE,
                 cwd=self.base_dir)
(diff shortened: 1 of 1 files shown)


Sheshuk:
Uh, so the time increased by 1 second. Here's a comparison from one of the previous runs:
 | Benchmark | Min | Max | Mean |
| :--- | :---: | :---: | :---: |
| python/snewpy/test/test_snowglobes.py::test_simulation_chain_benchmark| 4.34| 4.46| 4.40 +- 0.05 |

Sheshuk:
So we're having slightly decreased productivity because of the overhead of spawning threads wrt asyncio callbacks. But I think it's ok for now - it still is much more performant than it used to be, and we keep the code clean (without trying to hack around the Watchers in python 3.7). 
Also the snowglobes_interface will have to be updated when SNOwGLoBES v1.3 is out, probably we'll have more abilities to parallelize the jobs.

JostMigenda:
This looks good; thank you!

By the time SNOwGLoBES v1.3 is out, I think it’ll probably be okay to drop Python 3.7 support so we’ll have a bit more freedom then to choose the best path forward.

PR #159, opened 2021-11-30
refactoring models

Sheshuk:
Refactoring models: moving out the repeating code
reorganizing `models` -> `models.ccsn`


python/snewpy/models/base.py
```diff
@@ -0,0 +1,287 @@
...
+                    s += [f"|{name} | {v} |"]
+        return '\n'.join(s)
+
+    @abstractmethod
+    def get_time(self):
+        """Returns
+        -------
+            returns array of snapshot times from the simulation
+        """
+        pass
```
JostMigenda, line 62: Right now, this method is implemented in three different places:
* in `PinchedModel` (which most models in `snewpy.models.ccsn` inherit from; either directly or via `_GarchingArchiveModel`)
* and separately for the `Fornax_2019` and `Fornax_2021` models in `snewpy.models.ccsn`

In all three cases the implementation is simply `return self.time`. Would it make sense to move that to the base `SupernovaModel` class instead of having an abstract method here?

JostMigenda (changes requested):
Other than this comment on `get_time`, this looks good to me!

I’ll just add a commit with some minor cleanup stuff in a minute.

python/snewpy/models/base.py
Sheshuk: I was thinking about that.
It would mean that we require each subclass to set the `self.time` attribute, which is slightly less straightforward than requiring implementing the abstract method.

I mean if one doesn't implement the abstract method, a `TypeError` is raised on construction, but if the `self.time` is not defined, we'll get `AttributeError` only at the point we try to access it.
Is there any nice way to impose this requirement?

Sheshuk: Well, here's one possible solution. 
I can wrap the `__init__` method of the children classes on creation, to make the checks. It can be done with Metaclass, or, slightly simpler, with [`__init_subclass__`](https://docs.python.org/3/reference/datamodel.html#object.__init_subclass__) method:
```python
def check_model(self,*arg,**kwargs):
    if hasattr(self, 'time')==False:
        raise TypeError(f'{self.__class__}.time attribute not initialized in __init__!')
    
def wrap_init(init, check):
    @wraps(init)
    def _wrapper(self, *arg, **kwargs):
        init(self,*arg,**kwargs)
        check(self)
    return _wrapper

class SupernovaModel():
    def __init_subclass__(cls, **kwargs):
        super().__init_subclass__(**kwargs)
        cls.__init__ = wrap_init(cls.__init__, check_model)
```
Then having two models:
```python
class Model0(SupernovaModel):
    def __init__(self):
        self.time=[0,1,2,3]

class Model1(SupernovaModel):
    def __init__(self):
        pass
```
`Model0()` creates an instance, but `Model1()` raises `TypeError`.

Maybe it's not very important in general, but subclassing `SupernovaModel is one most possible way our code will be used. Also we can make additional checks for the models, if needed. What do you think?

Sheshuk: Implemented in 519d09f

Sheshuk:
@JostMigenda if you agree with the implementation of `get_time` we can merge this?

python/snewpy/models/base.py
JostMigenda: I get that it’s slightly nicer to raise an error on construction, rather than on first access; but am honestly not sure if it’s worth all this extra code. Since the target audience for this is developers who want to add their own module (who already need to read the code for the base class to figure out what functions to override), I’d be okay with letting this be a plain old AttributeError on access and just adding a comment (probably in the docstring of `SupernovaModel`’s `__init__`?) saying that the `__init__` function must set `self.time` and `self.metadata` and what their format should be.

Alternatively, I guess we could use `check_model()` to add some verification whether `self.metadata` is set & filled in?

Sheshuk: I'm also not sure if this is a good solution. Having abstract methods that user needs to implement seems cleaner. However if we switch to required attributes, I think it's important to have crash at the construction, otherwise user might start looking at the code in whatever downstream procedure tried to run `model.get_time`.
 
Also I agree that this is too much for a single check, and we can add more checks there.
For the `self.metadata` initialization - it was intended behavior, because I thought that in some cases no metadata can be provided by the model. But probably in that case one can set it explicitly to `{}`.

I also think the best way to make our intentions clear is to make `SupernovaModel.__init__(self, time, metadata)`which would fill the appropriate variables, so subclasses will call `super().__init__(times, metadata)` in their constructor.

JostMigenda: I think requiring the developer adding a model to think about metadata for a second is useful; the risk of them forgetting about it is larger than the very minor inconvenience of having to explicitly set it to `{}` in the very rare case where there actually isn’t any metadata to add.

The `super().__init__` way sounds cleaner; good idea!

Sheshuk:
Adding constructors and using them lead me to another thing: in most cases the model's constructors have a copy-paste part like this one: 
https://github.com/SNEWS2/snewpy/blob/4e49094d691e6bd747473ffcf7f7e930d295362c/python/snewpy/models/ccsn.py#L52-L66
I'll try to put it in a `PinchedModel._init_from_simtab` function`

Sheshuk:
Uh, looks like the latest release of astropy v5.0 doesn't support python 3.7. 
That's not critical - I just tried to use nice type annotations `Quantity['erg/(s*MeV)']` which makes the methods more self-documented. I'll roll back to 4b61733, so we're compatible again.
However this might be a problem, if we want to use astropy under python 3.7 in the future

Sheshuk:
@JostMigenda I think this PR is ready for review. Please have a look. I made constructors for both `SupernovaModel` and `PinchedModel` to reduce the repeating code. The `_post_init_check` is still present - in case we want to add some more sanity checks

JostMigenda (approved):
LGTM; thanks!

PR #161, opened 2021-12-13
remove duplicate times from Tamborra, Walk models

JostMigenda:
Input files for the `Tamborra_2014`, `Walk_2018` and `Walk_2019` models had multiple lines for certain times; either complete duplicates or with negligible numerical differences.

This trips up some algorithms (e.g. `scipy.interpolate.pchip` raises `ValueError: 'x' must be strictly increasing sequence.`), so this PR deletes those duplicate times.


JostMigenda:
Once this is merged into SNEWPY, JostMigenda/sntools#40 will add support for these models to sntools.

Sheshuk:
This looks like insufficient time precision in the simulation output script. 
Would it make sense to add such cleanup when we read the models, so if we later have more models from the Garching group, we don't have to manually clean them up?

JostMigenda:
I think I’d prefer it if a human takes a look at any such issue in the input file, as a sanity check; to confirm whether it’s actually just numerical noise like here or whether there’s a more serious issue.

Plus, if we publish model files it’s always gonna be possible that someone wants to use them independently of SNEWPY, so eliminating the problem at the root instead of fixing it in `snewpy.models` would be safer.

sybenzvi (changes requested):
All of these changes look fine except for `models/Tamborra_2014/s27.0c_3D_dir1_LS220_nux` where a large number of lines were deleted. Those do not appear to be duplicates so it's not obvious why this was done.

JostMigenda:
Ouch 😣 not sure how that happened; thanks for sorting it!
I have a meeting now but can fix that tomorrow.

sybenzvi (approved):
Thanks for restoring the deleted fluxes, this looks ready to go.

JostMigenda:
For future reference, here’s a quick script to remove duplicates:

```python

filename = 'models/Walk_2019/s40.0c_3DBH_dir1_LS220_nue'

fin = open(filename, 'r')
fout = open(filename+'_mod', 'w')

prev_t = -1

for line in fin:
    this_t = float(line.split()[0])
    if this_t <= prev_t:
        print(f"Deleted duplicate line at t={this_t}")
    else:
        fout.write(line)
    prev_t = this_t
```

PR #166, opened 2021-12-20
`Fornax_2021.get_initial_spectra()` performance improvements and bug fixes

JostMigenda:
Previously, `Fornax_2021`’s `__init__()` function would open the input file, read some data from it and then keep it open and read from it again in `get_initial_spectra()`. This PR changes this to read all required data from the file exactly once (in `__init__ ()`).

The result (measured by putting a `timeit.default_timer()` at the beginning/end of each function):
* `__init__()` stays at 20–30 ms (maybe a few % faster, but hard to say due to random fluctuation)
* `get_initial_spectra()` goes from ~13 ms to ~1 ms

Since `__init__()` only gets executed once, but `get_initial_spectra()` may get executed many times (e.g. once per time bin when generating SNOwGLoBES input files), this may make a significant difference in practice.

A nice side effect (and my main reason for looking into this in the first place): Keeping the open file as an instance variable meant that `Fornax_2021` objects couldn’t be [pickled](https://docs.python.org/3.10/library/pickle.html), which meant they couldn’t be used in a [multiprocessing](https://docs.python.org/3.10/library/multiprocessing.html?highlight=pickle) context, which e.g. sntools uses.


JostMigenda:
I’ve added a commit which returns number luminosity spectra (instead of luminosity spectra), like all other models. This fixes #167; see that issue for details.

Please note: For `interpolation='nearest'` I think I’ve discovered a bug in the treatment of energy bins—a simple mix-up between bin centres and edges. I’ve fixed that in a separate commit and made the variable names and comments a little clearer.

@sybenzvi, since you originally added the Fornax models, could you review this PR?

sybenzvi (approved):

sybenzvi:
I'm fine with these changes so I'll merge the PR.

PR #168, opened 2022-01-05
Removed duplicate times from Bollig_2016 s27.0c files

svalder:
Similar to PR #161 the `s27.0c` files in `Bollig_2016` had multiple lines with the exact same time stamp. This in turn causes an error in `scipy.interpolate.pchip`. 

This PR deletes these lines following the resolution in PR #161. 

Note: the mean energy are very similar but the luminosities in each example appear higher.


sybenzvi (approved):

JostMigenda:
Hm … unit tests all succeed and the integration tests themselves also succeed; the failure occurs in the last step (posting benchmark results here) and is unrelated to this change. (Though unfortunately, the error message is vague enough that I can’t tell what it _is_ related to.) So I’d suggest we go ahead, merge this PR and then wait until the next PR to see whether this is a temporary issue and resolves itself or whether we need to fix the integration test workflow?

sybenzvi:
> 

I agree @JostMigenda, I can't figure out why the integration test fails and there is no good reason why this change is at fault. I will try to merge the branch anyway.

PR #171, opened 2022-01-12
Added a dictionary of keyword arguments for SN models + added code for ROOT output

soso128:
This pull request contains two commit:

1- Added an argument to generate_time_series and generate_fluence in snowglobes.py: this dictionary allows to give keyword arguments to define SN models (EOS, etc...)

2- Added a new code to get ROOT 2D histograms for the rates, instead of collated text files

### DIFF
--- a/python/snewpy/snowglobes.py
+++ b/python/snewpy/snowglobes.py
@@ -40,7 +40,7 @@ from snewpy.snowglobes_interface import SNOwGLoBES
 
 logger = logging.getLogger(__name__)
 
-def generate_time_series(model_path, model_type, transformation_type, d, output_filename=None, ntbins=30, deltat=None):
+def generate_time_series(model_path, model_type, transformation_type, d, output_filename=None, ntbins=30, deltat=None, snmodel_dict={}):
     """Generate time series files in SNOwGLoBES format.
 
     This version will subsample the times in a supernova model, produce energy
@@ -75,7 +75,7 @@ def generate_time_series(model_path, model_type, transformation_type, d, output_
     flavor_transformation = flavor_transformation_dict[transformation_type]
 
     model_dir, model_file = os.path.split(os.path.abspath(model_path))
-    snmodel = model_class(model_path)
+    snmodel = model_class(model_path, **snmodel_dict)
 
     # Subsample the model time. Default to 30 time slices.
     tmin = snmodel.get_time()[0]
@@ -144,7 +144,7 @@ def generate_time_series(model_path, model_type, transformation_type, d, output_
     return os.path.join(model_dir, tfname)
 
 
-def generate_fluence(model_path, model_type, transformation_type, d, output_filename=None, tstart=None, tend=None):
+def generate_fluence(model_path, model_type, transformation_type, d, output_filename=None, tstart=None, tend=None, snmodel_dict={}):
     """Generate fluence files in SNOwGLoBES format.
 
     This version will subsample the times in a supernova model, produce energy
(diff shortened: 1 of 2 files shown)


python/snewpy/snowglobes.py
```diff
@@ -40,7 +40,7 @@
 
 logger = logging.getLogger(__name__)
 
-def generate_time_series(model_path, model_type, transformation_type, d, output_filename=None, ntbins=30, deltat=None):
+def generate_time_series(model_path, model_type, transformation_type, d, output_filename=None, ntbins=30, deltat=None, snmodel_dict={}):
```
JostMigenda, line 43: Could you add a description of this new parameter do the docstring below, please? E.g., something like
```
snmodel_dict : dict
    Additional arguments for setting up the supernova model. See documentation of relevant ``SupernovaModel`` subclass for available options. (Optional.)
```
(And same in `generate_fluence()` below.)

JostMigenda (changes requested):

python/snewpy/snowglobes.py
soso128: Done!

JostMigenda:
Thanks @soso128!

I see you’ve already added the docstrings—excellent! Unfortunately, it looks like you accidentally included some unrelated changes to `collate()` in 3bc87a4. Once you remove those,  I think part 1 looks good.

Regarding part 2: ROOT is a very heavy dependency (in terms of time and disk space to install) and, in my experience, quite fragile when it comes to binary compatibility with different versions of Python. So I would be _very_ hesitant before depending on it for a core feature of SNEWPY. Since this code also appears to duplicate a lot of the code from `snewpy.snowglobes.collate()`, maybe it would be better to have a sample script in `snewpy/python/scripts/` that demonstrates how to take the tables returned by `snewpy.snowglobes.collate()` and write those to a ROOT file?

soso128:
Ok, done. Indeed it would be better to write a code in scripts/ for the ROOT output. I'll do this.

soso128:
I just added a script for the ROOT output (and removed the previous code)

python/snewpy/scripts/Convert_to_ROOT.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python
+""" This module allows to gather the SNOwGLoBES outputs into a ROOT files with 2D histograms.
+    These 2D histograms show the rate as a function of time and energy for any given interaction channel.
+    There is one ROOT file per detector, per smearing option, and per weighting option.
+
+    To run the code, use:
```
JostMigenda, line 6: This script currently only works for tarballs generated by `generate_time_series()`, not `generate_fluence()` where the filenames are a bit different. For a sample script, I think that’s okay; but we should add a brief explanation at the start to avoid confusion.
```suggestion
    Generate a tarball with the input files by running
      snowglobes.generate_time_series(...)
      snowglobes.simulate(...)
      snowglobes.collate(...)

    Then, to run the code, use:
```

JostMigenda (changes requested):
Thanks for moving this to a script! It looks good, I just have a few minor suggestions to make it a little clearer to the user.

python/snewpy/scripts/Convert_to_ROOT.py
soso128: I implemented the changes

JostMigenda (approved):
All looks good now; many thanks @soso128 for being so responsive! And congratulations on your first contribution to SNEWPY! 🎉

PR #172, opened 2022-01-12
Add rate computation without GLOBES

soso128:
Since someone in my group didn't want to install GLoBES, I modified the Snowglobes interface to allow to compute the expected rates using only the SNOwGLoBES detector configurations and cross-sections. Here, the rates are computed for a perfect detector (no smearing, 100% efficiencies). I added options in snowglobles.simulate and snowglobes.collate in order to choose to compute this rate if needed.
Running this code for IceCUBE and for the Nakazato 2013 model with 30 bins (default options in scripts/TimeSeries.py), I found that the discrepancies between the new and old unsmeared rates were of at most 0.4%.

### DIFF
--- a/python/snewpy/snowglobes_interface.py
+++ b/python/snewpy/snowglobes_interface.py
@@ -218,6 +218,7 @@ class Runner:
     def __post_init__(self):
         self.channels=self.sng.channels[self.material]
         self.binning=self.sng.binning[self.material]
+        self.efficiency=self.sng.efficiencies[self.detector]
         self.det_config=self.sng.detectors[self.detector]
         self.base_dir=self.sng.base_dir
         self.out_dir=self.base_dir/'out'
@@ -283,9 +284,14 @@ class Runner:
         else:
             raise RuntimeError('SNOwGLoBES run failed:\n'+stderr)
 
-class SimpleRate(SNOwGLoBES):
+class SimpleRate():
     def __init__(self, base_dir:Path=''):
-        """ SNOwGLoBES interface 
+        """ Simple rate calculation interface 
+        Computes expected rate for a perfect detector (100% efficiencies, no smearing)
+        without using GLOBES. The formula for the rate is
+                Rate = [cross-section in 10^-38 cm^2] x 10^-38 x [fluence in cm^-2] x [target mass in kton] 
+                    x [Dalton per kton] x [energy bin size in GeV]
+        with [target mass in kton] x [Dalton per kton] = number of reference targets in experiment.
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


Sheshuk (commented):
It's nice to have a separate rates calculation! I think it should also be possible to include also the smearing effects from SNOwGLoBES - because basically GLoBES is used to multiply vector by matrix and for interpolation. 

However I think this is strictly speaking not SNOwGLoBES. So I suggest to separate this functionality in a new class (perhaps a child of `SNOwGLoBES`) - it saves a code several `if`s? will be less misleading and also will not complicate the SNOwGLoBES class usage with additional flags

Sheshuk (changes requested):
Also, when doing rate calculations it doesn't make sense to run in parallel threads (because python will do them sequentially anyway), so I suggest to:
1. Subclass `SNOwGLoBES`
2. Reimplement `run(self, flux_files, detector, material)` method using your code from `_compute_rates`.

soso128:
Hi, thank you for the suggestion! I now implemented a separate class with its own run function.

python/snewpy/snowglobes_interface.py
```diff
@@ -218,8 +218,6 @@ class Runner:
     def __post_init__(self):
         self.channels=self.sng.channels[self.material]
         self.binning=self.sng.binning[self.material]
-        self.efficiency=self.sng.efficiencies[self.detector]
```
Sheshuk, line 221: Is there a reason for removing `self.efficiency` and `self.smearing` initialization? 
Probably they need to be restored, for the Runner class to work

python/snewpy/snowglobes_interface.py
```diff
@@ -284,5 +282,109 @@ def run(self):
             return tables
         else:
             raise RuntimeError('SNOwGLoBES run failed:\n'+stderr)
+
+class SimpleRate(SNOwGLoBES):
+    def __init__(self, base_dir:Path=''):
+        """ SNOwGLoBES interface
```
Sheshuk, line 288: Could you put the description of the rate calculation in the docstring here?

python/snewpy/snowglobes_interface.py
```diff
@@ -284,5 +282,109 @@ def run(self):
             return tables
         else:
             raise RuntimeError('SNOwGLoBES run failed:\n'+stderr)
+
+class SimpleRate(SNOwGLoBES):
```
Sheshuk, line 286: Oh, if you don't reuse the `__init__` method from `SNOwGLoBES`, there is no reason to subclass SNOwGLoBES Sorry, I didn't realize in the first place, when I suggested you to use subclass. In the current implementation, these two classes don't have any shared methods, so subclassing has no effect.

Sheshuk (changes requested):
Thanks for implementing my suggestions, this looks good! If you address the comments, I think we'll be ready to merge.

python/snewpy/snowglobes_interface.py
soso128: Actually I am using the _load_detectors and _load_channels methods of the SNOwGLoBES  class. Do you think I should just add them to the SimpleRate class and drop the inheritance?

python/snewpy/snowglobes_interface.py
soso128: This was an accident, I'll add it back

python/snewpy/snowglobes_interface.py
Sheshuk: You're right, sorry, I missed it. Then it's good and completely adequate as it is

soso128:
I have put the "self.efficiency=self.sng.efficiencies[self.detector]" line back and moved the docstring

Sheshuk (approved):
Looks good! Thanks for implementing this

Sheshuk:
Oops, I overlooked that `Runner.smearing` is still missing...

soso128:
Oops, sorry, I have just put it back

Sheshuk:
Too late - I already merged it :( Created #174 with a hotfix

soso128:
Ok, thanks a lot and sorry for the mistake!

PR #173, opened 2022-01-12
`snewpy.snowglobes`: note deprecated parameters

JostMigenda:
With the rewrite of the SNOwGLoBES interface in #132, several parameters to `simulate()` and `collate()` are no longer used. To avoid confusing users, this PR updates the docstrings to note that these parameters are deprecated and prints a warning message if one of these parameters is used.

### DIFF
--- a/python/snewpy/snowglobes.py
+++ b/python/snewpy/snowglobes.py
@@ -27,11 +27,11 @@ import tarfile
 from pathlib import Path
 from tempfile import TemporaryDirectory
 
-import matplotlib as mpl
 import matplotlib.pyplot as plt
 import numpy as np
 from astropy import units as u
 from tqdm.auto import tqdm
+from warnings import warn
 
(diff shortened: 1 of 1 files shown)


Sheshuk (commented):
Looks good! Just a thought for the future work: we could use `logging.warning` instead of `warn`, to be more flexible and consistent about the logging.

Sheshuk (approved):

JostMigenda:
Regarding `logging.warning()` vs. `warnings.warn()`: [Documentation recommends](https://docs.python.org/3/howto/logging.html#when-to-use-logging) the latter for this purpose. If I understand correctly, the difference is that the former only prints text to stdout (or a file, or however logging is configured), whereas the latter raises an actual `Warning` (which may get printed to stdout or stderr by default; but gives the users more options to configure warning filters, e.g. to treat all warnings as errors).

PR #176, opened 2022-01-18
Bug fix in SimpleRate

soso128:
Hi,
Sorry I inadvertently removed the Snowglobes inheritance in the SimpleRate class. This commit adds it back.


soso128:
I also fixed a but in the Convert_to_ROOT.py script: the script worked fine for a single detector but didn't update interaction channels properly when processing multiple detectors.

JostMigenda:
Hi @soso128, thanks for fixing this so quickly! Two things:

1. Can you take a look at https://github.com/soso128/snewpy/blob/soso128_test/python/snewpy/snowglobes_interface.py#L300 – I think a few words accidentally got deleted there; it should probably say “Path to the SNOwGLoBES installation”, just like in the base class?
2. Could we add a few tests for this, to ensure we immediately notice if this part of the code ever breaks again? (See `python/snewpy/test/test_snowglobes.py` and `…/snowglobes_integrationtest.py` for similar tests that could be adapted.)

soso128:
Hi, thank you for your feedback! I fixed the docstring and added a code to the test suite (is it automatically accounted for? I didn't see any test file list so I assumed so)

soso128:
Ok, after a few issues the tests are now incorporated and seem to run fine.

python/snewpy/test/test_snowglobes.py
```diff
@@ -30,18 +30,39 @@ def test_generate_fluence(sng,model,transformation):
...
+                    ('halo2',                 146.4671)]
+
 
 @pytest.fixture(autouse=True)
 def sng():
     return SNOwGLoBES()
 
+@pytest.fixture(autouse=True)
+def splr():
+    return SimpleRate()
```
JostMigenda, line 50: Remember to import `SimpleRate` at the top of the file!

JostMigenda:
Excellent, thanks for the changes!

Regarding the test suite, SNEWPY has two separate sets of tests: unit tests (which run automatically on every commit) and integration tests (which only run when manually triggered, e.g. by requesting a review as I did just now).

Both unit and integration tests automatically include all tests in files which follow the conventional `test_*.py` file name pattern. The `simplerate_integrationtest.py` which you added doesn’t match that (like the previously added `snowglobes_integrationtest.py`), so you need to manually add it in https://github.com/SNEWS2/snewpy/blob/main/.github/workflows/integration.yml#L58. (Note that GitHub is a bit restrictive when it comes to editing workflows; if you run into issues, please try it in the web editor.)

One final question: I notice that the number of events in [the two crosscheck tables](https://github.com/SNEWS2/snewpy/pull/176/files#diff-3ba8a323e58bd9db6e85a6bf5df8d70ed7945ec3d23d36211cda272380119a2eR25-R42) is within rounding error for most detectors, but differs by 3× for HALO and over 20× for IceCube. What’s going on there? Is that simply because the difference between smeared and unsmeared event numbers is unusually large for these detectors? Or is there something more complicated going on?

soso128:
Hi Jost,

Thanks for your reply, I edited the workflow (which got associated with its own pull request it seems) and added the import for test_snowglobes.py.
I was also very surprised to find such large discrepancies between the smeared and unsmeared rates for some experiments. I verified that, using my scripts, the smeared rates corresponded to the ones quoted in the test code, which is why I ended up trusting the unsmeared rates. But yes, I am still very surprised by such a large discrepancy. For IceCube it might be due to the fact that the energy smearing is actually huge.

python/snewpy/test/test_snowglobes.py
```diff
@@ -30,18 +30,39 @@ def test_generate_fluence(sng,model,transformation):
...
     data = sng.run(flux,detector)
     total = data[0].weighted.smeared.sum().sum()
     assert total == pytest.approx(expected_total, 0.1)
 
+@pytest.mark.parametrize('detector, expected_total_unsmeared',crosscheck_table_unsmeared)
+def test_simplerate_crosscheck(splr, detector, expected_total_unsmeared):
+    flux = './models/Bollig_2016/fluence_Bollig_2016_s11.2c_AdiabaticMSW_NMO.dat'
+    data = splr.run(flux,detector)
+    total = data[0].weighted.unsmeared.sum().sum()
+    assert total == pytest.approx(expected_total, 0.01)
```
JostMigenda, line 64: Another tiny bug that causes a few of the integration tests to fail:
```suggestion
    assert total == pytest.approx(expected_total_unsmeared, 0.01)
```

python/snewpy/test/simplerate_integrationtest.py
```diff
@@ -0,0 +1,57 @@
...
+        tarredfile = snowglobes.generate_fluence(model_path, modeltype, transformation, distance, outfile)
+
+        print("Simulating detector effects with SNOwGLoBES ...")
+        snowglobes.simulate(SNOwGLoBES_path, tarredfile, detector_input=detector, detector_effects=False)
+
+        print("Collating results ...")
+        tables = snowglobes.collate(SNOwGLoBES_path, tarredfile, detector_input=detector, skip_plots=True, smearing=False)
+
+        # Use results to print the number of events in different interaction channels
+        key = f"Collated_{outfile}_{detector}_events_smeared_weighted.dat"
```
JostMigenda, line 39: Use unsmeared events for `SimpleRate` test
```suggestion
        key = f"Collated_{outfile}_{detector}_events_unsmeared_weighted.dat"
```

JostMigenda (approved):
Just fixed two more tiny copy&paste errors in the integration tests. They run successfully now[*] and I ran the one missing test (`simplerate_integrationtest.py`) on my local machine to verify that it works, too. I will merge #178, which adds that one to the integration tests workflow, in a moment.

[*] The workflow does fail, but only a later stage, after the integration tests have all succeeded. This also happened in #168; I’m wondering whether it is due to the origin branch being in a forked repo …

soso128:
Sorry for having given you more work and thanks for fixing the remaining issues!

JostMigenda:
Don’t worry, it happens. (Maybe that’s a good reminder to run tests locally, too—the GitHub Actions are an important safety net, but running tests locally makes it much quicker to discover and fix these tiny bugs.)

PR #178, opened 2022-01-20
Update integration.yml to include SimpleRate

soso128:
Add integration test for SimpleRate


JostMigenda (approved):
As discussed in #176.

PR #181, opened 2022-01-24
Added badge with latest release to README

Sheshuk:


README.md
```diff
@@ -1,6 +1,7 @@
 # SNEWPY: Supernova Neutrino Early Warning Models for Python
 
 [![DOI](https://zenodo.org/badge/221705586.svg)](https://zenodo.org/badge/latestdoi/221705586)
+![Release](https://img.shields.io/github/v/release/SNEWS2/snewpy?include_prereleases)
```
JostMigenda, line 4: Good idea; though I’d suggest linking to PyPI where users can also download the latest release:
```suggestion
[![PyPI](https://img.shields.io/pypi/v/snewpy)](https://pypi.org/project/snewpy/)
```

(The displayed version would be the same in both cases, since we’re publishing all new releases to PyPI automatically.)

JostMigenda (approved):

PR #194, opened 2022-07-21
Require AstroPy >=4.3

### LINKED ISSUE #188: Loading `Kuroda_2020` model fails with certain versions of astropy
Attempting to initialize the `Kuroda_2020` model with code like the following, while using astropy v4.2.1 yields a rather uninformative error. This error seems to involve a data column that is expected to have astropy units, but doesn't when the model file is actually loaded. This same error is thrown for all three `Kuroda_2020` models. Note that the line numbers displayed below may not correspond to the main branch, but do correspond to `Kuroda_2020.__init__`.

I will note, running the same code in astropy 4.3.1 yields no errors. Would it be appropriate/sufficient to add a version requirement (`astropy >= 4.3.1`) to the `requirements.txt`?

Example Code:
```python
>>> from snewpy.models.ccsn import Kuroda_2020
>>> Kuroda_2020('/path/to/snewpy/models/Kuroda_2020/LnuR00B00.dat')
```
Error:
```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_7249/1500263364.py in <module>
      1 from snewpy.models.ccsn import Kuroda_2020
----> 2 Kuroda_2020('/path/to/snewpy/models/Kuroda_2020/LnuR00B00.dat')

~/Code/SNEWS/snewpy/python/snewpy/models/base.py in _wrapper(self, *arg, **kwargs)
     16     @wraps(init)
     17     def _wrapper(self, *arg, **kwargs):
---> 18         init(self, *arg, **kwargs)
     19         check(self)
     20     return _wrapper

~/Code/SNEWS/snewpy/python/snewpy/models/ccsn.py in __init__(self, filename, eos, mass)
    433             simtab[f'E_{f.name}'] = simtab[f'<E{fkey}>'] << u.MeV
    434             # There is no pinch parameter so use alpha=2.0.
--> 435             simtab[f'ALPHA_{f.name}'] = np.full_like(simtab[f'E_{f.name}'].value, 2.)
    436 
    437         super().__init__(simtab, metadata)

AttributeError: 'Column' object has no attribute 'value'
```

JostMigenda:
Fixes #188. AstroPy 4.3 still supports Python 3.7 or higher (same as SNEWPY), so this won’t affect backwards compatibility.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 numpy >= 1.16
 scipy
-astropy
+astropy >=4.3
 pandas == 1.3.4
 Jinja2==3.0.2
 tqdm


sybenzvi (approved):
I think this should be alright, given that the current stable release is astropy 5.1.

PR #196, opened 2022-08-01
extend .gitignore list

JostMigenda:
Ignore output files from running `snewpy.snowglobes` as well as files generated during integration tests in `test_snowglobes.py`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,13 @@ dist/
 # default directory for downloaded model files
 SNEWPY_models/
 
+# snewpy.snowglobes output files
+*_SNOprocessed.tar.gz
+*.npy
+
+# used for integration tests in test_snowglobes.py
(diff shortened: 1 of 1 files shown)


sybenzvi (commented):
Looks reasonable, but instead of ignoring just

`fluence_Bollig_2016_s*`

maybe it should ignore all possible fluence files? E.g.,

`fluence_s*`

JostMigenda:
I searched the whole repo for `fluence_` and it looks like filenames with starting with that string are only generated for the Bollig model. So right now there’s no need to be more general; and I worry that ignoring all `fluence_*` files might be too broad; in case we get e.g. some `fluence_calculation.py` file in the future.

sybenzvi (approved):
OK, I think this can proceed.

PR #198, opened 2022-08-02
add `MixingParameters` presets

### LINKED ISSUE #195: `MixingParameters` presets
Currently, the `snewpy.neutrino.MixingParameters` class contains two sets of default oscillation parameters (for NMO and IMO), both from NuFit 5.0 (July 2020). Since then, there’s been a 5.1 update in October 2021 and NuFit 6.0 is likely to come out later this year. We should regularly update these parameters as new NuFit (or PDG?) publications come out.

A problem with updating those parameters is that different SNEWPY versions would then give different results. This prevents users from updating and may also break some automated tests. To eliminate these issues, I would suggest adding a second argument to the initialisation, so that we could create MixingParameters e.g. using `MixingParameters(MassHierarchy.INVERTED, "NuFit5.0")` or `MixingParameters(MassHierarchy.NORMAL, "NuFit5.1")`.
Calling it without this second argument, i.e. just `MixingParameters(MassHierarchy.NORMAL)`, would always refer to the latest data.

JostMigenda:
Closes #195.
In addition to the NuFIT5.0 values, this PR adds NuFIT5.1 and PDG2022 mixing parameter values and an optional argument to select which version of the parameters to use. In the initial commit, the default remains `NuFIT5.0`. In an upcoming commit, I will explicitly set the version in any tests that depend on this, so tests don’t start to fail once we change the default to a more recent version of the parameters.


JostMigenda:
The test failure is fixed in #207.

Sheshuk (changes requested):
Adding a version in the parameter into the constructor doesn't seem right to me. This way we'll have to expand the definition each time new version is available, changing the class.

What I think would be cleaner is:
1. Use the [`dataclass`](https://docs.python.org/3/library/dataclasses.html) decorator (`MixingParameters` class is basically a data container). It will provide a nice constructor, where wwe can pass all the values separately.
2. For each NUFIT version make the objects of the `MixingParameters` class, so it would be available when you import these parameters, something like:
```python
from snewpy.neutrino import pars_NuFIT_5_2
```
3. Pass these parameters as object to where they are used (`FlavorTransformations`)

If you agree, I can commit this implementation

JostMigenda:
Re: 3., see [my comment here](https://github.com/SNEWS2/snewpy/pull/189#issuecomment-1553400690). But I think that’s just a slight annoyance and not a reason to avoid dataclasses; so yes, please go ahead!

Sheshuk:
Um, in the mixing parameters we have `dm32_2` for NH and `dm31_2` for IH, which is correct, but makes it hard to use with the dataclasses.

I did `grep dm3 *.py` in the code and looks like we don't actually use any of them.
It's rather hard to use them directly: one would have to make checks on hierarchy and choose the right value.

So maybe we can just name them `dm2_small` and `dm2_large` or something similar?

Another option would be to have separate classes for NH and IH, but I'd prefer otherwise

Sheshuk:
> So maybe we can just name them `dm2_small` and `dm2_large` or something similar?

Or `dm2_21` and `dm2_3l` as it is in NuFIT

JostMigenda:
> Um, in the mixing parameters we have `dm32_2` for NH and `dm31_2` for IH, which is correct, but makes it hard to use with the dataclasses.

Careful, it’s even trickier: NuFIT gives `dm32_2` and `dm31_2` depending on the mass ordering, while PDG gives `dm32_2` for both NMO and IMO.

I guess in principle, if we know `dm31_2` and `dm21_2` for NuFIT IMO, we can calculate `dm32_2` from that and use that for all cases? That would certainly be easiest to implement in this PR; and if they’re not used anywhere right now, then it can’t cause a problem elsewhere.
But just to be sure—@jpkneller, do you see that causing any issues when implementing future `FlavorTransformation`s?

JostMigenda:
> Or `dm2_21` and `dm2_3l` as it is in NuFIT

Since `l` and `1` look very similar in many fonts, I’d really rather avoid that! (The NuFIT folks get around that by using the cursive `𝓁`, but we can’t do that in code.)

jpkneller:
If I have to pick, I would prefer dm32_2 over dm31_2. Even better would be to have all three available although that might open the possibility that a user could assign values that were not self consistent.

Sheshuk:
I implemented the class to hold all dm_2, and calculate the missing ones (passed as None), also I added validation after the constructor, that the proper sum of dm_2 is zero.

@JostMigenda in your presets I see values of **dm32** for **NH** and **dm31** for **IH**, but I suppose it should be vice versa? Because on [NuFIT](http://www.nu-fit.org/?q=node/256) site I read:
```
Note that Δm​2_3ℓ = Δm​2_31 > 0 for NO and Δm​2_3ℓ = Δm​2_32 < 0 for IO. 
```
So probably I should correct this in the test?

python/snewpy/neutrino.py
```diff
@@ -37,50 +47,65 @@ def is_neutrino(self):
...
-            self.theta13 =  8.60 * u.deg
-            self.theta23 = 49.30 * u.deg
-            self.deltaCP = 282 * u.deg
-            self.dm21_2  = 7.42e-5 * u.eV**2
-            self.dm31_2  = -2.498e-3 * u.eV**2
+            assert self.dm32_2<0, 'dm32_2 should be negative for IH'
+            assert self.dm31_2<0, 'dm31_2 should be negative for IH'
+        #validate dm2
+        dm2_sum = self.dm32_2+self.dm21_2-self.dm31_2
+        assert dm2_sum==0, f'dm32_2+dm31_2-dm31_2 = {dm2_sum} !=0'
```
JostMigenda, line 108: Is there a risk of running into numerical issues here? (E.g. note that `assert 0.1 + 0.1 + 0.1 == 0.3` raises an `AssertionError`.)

JostMigenda:
Good catch about the 31/32 mixup, thanks!

I think the `MixingParameters4Flavor` class should have another mass-squared difference for the 4th neutrino flavour. (Or rather three, of which one needs to be provided, and two can be calculated automatically?)

python/snewpy/neutrino.py
Sheshuk: You're right, thanks. I'll fix it with `np.isclose`

Sheshuk:
Okay, I've added the needed parameters.
Also I made it possible to create `MixingParameters4Flavor` like this, to support extending standard 3 flavor ones:
```python
pars_3f = MixingParameters() #standard 3flavor mixing
pars_4f = MixingParameters4Flavor(**pars_3f, theta14=90<<u.deg, dm41_2=1<<u.GeV**2)
```

python/snewpy/neutrino.py
```diff
@@ -91,4 +124,111 @@ def get_mixing_angles(self):
...
+        super().__post_init__()
+        self.dm42_2 = self.dm42_2 or self.dm41_2 + self.dm21_2
+        self.dm43_2 = self.dm43_2 or self.dm41_2 + self.dm31_2
+
+        dm2_sum = self.dm41_2 - self.dm42_2 + self.dm21_2
+        assert np.isclose(dm2_sum,0), f'dm41_2-dm42_2-dm21_2 = {dm2_sum} !=0'
+        dm2_sum = self.dm41_2 - self.dm43_2 + self.dm31_2
+        assert np.isclose(dm2_sum,0), f'dm41_2-dm43_2-dm31_2 = {dm2_sum} !=0'
+        dm2_sum = self.dm42_2 - self.dm43_2 + self.dm32_2
+        assert np.isclose(dm2_sum,0), f'dm42_2-dm43_2-dm32_2 = {dm2_sum} !=0'
```
JostMigenda, line 165:
```suggestion
        self.dm42_2 = self.dm42_2 or self.dm41_2 - self.dm21_2
        self.dm43_2 = self.dm43_2 or self.dm41_2 - self.dm31_2

        dm2_sum = self.dm41_2 - self.dm42_2 - self.dm21_2
        assert np.isclose(dm2_sum,0), f'dm41_2 - dm42_2 - dm21_2 = {dm2_sum} !=0'
        dm2_sum = self.dm41_2 - self.dm43_2 - self.dm31_2
        assert np.isclose(dm2_sum,0), f'dm41_2 - dm43_2 - dm31_2 = {dm2_sum} !=0'
```
Please double-check—it’s so easy to mix up the signs!—but I think they should be correct like this.

Also, I would delete the final assert—if the first two asserts here plus the assert in the `super().__post_init__` are all true, that final assert is automatically true.

python/snewpy/neutrino.py
```diff
@@ -91,4 +124,109 @@ def get_mixing_angles(self):
...
+    ..Note: it is an extension of :class:`MixingParameters3Flavor`, and can be constructed using it:
+    
+        >>> pars_3f = MixingParameters() #standard 3flavor mixing
+        >>> pars_4f = MixingParameters4Flavor(**pars_3f, theta14=90<<u.deg, dm41_2=1<<u.GeV**2)
+    """
+    #sterile neutrino miging angles
+    theta14: u.Quantity[u.deg] = 0<<u.deg
+    theta24: u.Quantity[u.deg] = 0<<u.deg
+    theta34: u.Quantity[u.deg] = 0<<u.deg
+    #sterile neutrino mass squared differences
```
mcolomermolla, line 150: why setting them to zero instead of being an input parameter the user can give?

mcolomermolla (commented):
I like the new way to handle the versions for the oscillation parameters!
Minor comment: is there a reason to have NuFit5.0 and 5.2 but not 5.1 version?

python/snewpy/neutrino.py
JostMigenda: Note that this is marked as a [dataclass](https://docs.python.org/3.11/library/dataclasses.html), which automatically generates a `__init__` function to set them as input parameters. The value defined here is used as the default.

JostMigenda:
We keep 5.0 as the initial value (for backwards compatibility) and 5.2 as the most up-to-date value. I don’t think there is a reason to newly add an outdated version.

Sheshuk (dismissed):

python/snewpy/neutrino.py
Sheshuk: Yes, as Jost said zeros will be default parameters in the constructor. So by default there is no mixing with sterile neutrinos, and only if user provides the mixing angles in the constructor, e.g.:
```python
params4f = MixingParameters4Flavor(**params3f, theta24=45<<u.deg)
```
there will be an effect of sterile neutrinos

mcolomermolla (approved):
Ready to be merged

Issue #183, opened 2022-02-04
Create a method to compose flavor transformations


sybenzvi:
Do we want snewpy to be capable of stacking a set of flavor transformations? For example, given a model we apply some resonant effect in the star, followed by vacuum oscillations, followed by MSW Earth-crossing oscillations.

Currently the interface to implement transformations is `get_transformed_spectra(self, t, E, flavor_xform)`, so we could take the existing flavor transformation classes and setup a way to compose several different processes into one object that gets passed to `get_transformed_spectra`.

Issue #99, opened 2021-09-11
Treat μ and τ flavours separately


JostMigenda:
Currently, SNEWPY does not distinguish between μ and τ neutrinos, referring to both as `NU_X`. (Same for μ and τ antineutrinos, referred to as `NU_X_BAR`.) While this is in line with most computer models, there are a few that treat μ and τ separately and we should consider updating SNEWPY to deal with this.

This would require changes to the `Flavor` enum in `snewpy.neutrino`, as well as to all flavor transformations. There will also be performance downsides of having 6 flavours instead of 4—we should see how large that effect is in practice and whether it’s possible to avoid it for models that don’t benefit from the 6-flavour treatment.

sybenzvi:
Do you have a list of models where this will matter? And how much do we lose by merging mu and tau flavored fluxes into a nu_x flux for models that use 3 flavors instead of nu_e + nu_x + antineutrinos?

JostMigenda:
Figure 3 of [arXiv:1706.04630](https://arxiv.org/pdf/1706.04630.pdf) indicates a roughly ~10% difference in luminosity between mu and tau antineutrinos if muon production is included in simulations. (Plus overall higher luminosity & mean energy as well as earlier explosions, compared to simulations that ignore this effect.)
I don’t know any others off the top of my head (apart from [arXiv:2008.13628](https://arxiv.org/pdf/2008.13628.pdf) fig. 9, which covers just 30 ms around the core bounce), but maybe @evanoconnor can help?

evanoconnor:
right, so the models will be different.  But maybe what @sybenzvi is asking is if we just lump all heavies together into nux, what difference would _we_ see.  Currently, I think not much.  SNOwGLoBES, for example, doesn't have the charged current muon interactions, so distinguishing mu and tau doesn't do anything (@jpkneller can correct me here, but I think with a six flavour mixing any differences between mu and tau neutrinos show up in the electron flavour at the sin^2(theta_{13}) level, or something like that).  the neutrino - antineutrino difference is bigger as when rolled in with oscillations (now the differences come in depending on theta_{12}) it can impact the electron flavor at earth.

JostMigenda:
[...]

PR #203, opened 2022-08-04
First pass at composite transformations.

sybenzvi:
This PR is designed to address issue #183 by implementing a simple class to compose an arbitrary list of flavor transformations.

If instantiated as follows,
`c = CompositeTransformation(U1, U2, ..., Un)`
where `U1`, `U2`, etc. are different `FlavorTransformation` objects, the class will apply the transformations in the order supplied to the CompositeTransformation. I.e., `U1` will be applied first, then `U2`, and so on to `Un`.

Unit tests need to be added to the branch before we proceed with the merger.

### DIFF
--- a/python/snewpy/flavor_transformation.py
+++ b/python/snewpy/flavor_transformation.py
@@ -162,6 +162,201 @@ class FlavorTransformation(ABC):
         pass
 
 
+class CompositeTransformation(FlavorTransformation):
+    """Composite surival probabilities for a set of transformations.
+    """
+
+    def __init__(self, *xforms):
+        """Initialize with an arbitrary list of FlavorTransformation objects.
+        The transformations are applied in the order they are provided.
+
+        Parameters
+        ----------
+        *xforms : FlavorTransformation
+            A variable argument list of FlavorTransformation objects.
+        """
+
+        # Store a list of transformations. The first transformation in the list
+        # is applied first, the second is applied second, etc.
+        self._xforms = xforms
+
+        # Composed transformation matrices for neutrinos and antineutrinos.
+        self._U_nu  = np.identity(2, dtype=float)
+        self._U_bar = np.identity(2, dtype=float)
+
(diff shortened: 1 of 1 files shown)


sybenzvi:
Comment about the guts of `CompositeTransformation`: internally it performs matrix multiplication of the composed unitary transformations. The composite matrix is not supposed to be accessed by users, who should use the survival and transition probability functions `prob_ee(t, E)` etc. provided by the base class. For efficiency, the calculation of the transformation matrix is memoized so that it is not recalculated for all transition probabilities unless the input time and energy arguments change.

On this topic, the following checks should be completed:
- [ ] Unit test of the interface
- [ ] Checks that memoization works as intended
- [ ] Checks that ndarray arguments for input time and energy work

JostMigenda:
At first glance, this looks good to me; thanks! I would also suggest updating the `FlavorTransformation.ipynb` to add an example for this composite transformation.

jpkneller:
I checked the math and it turns out chaining together SNEWPY's flavor transformations cannot be done even if the previously mentioned restriction that the neutrinos decohere between transformations is valid. Attached is a proof. 
[proof.pdf](https://github.com/SNEWS2/snewpy/files/11630696/proof.pdf)

JostMigenda:
Thanks for checking this and writing it up! If I understand correctly, there’s two different ways we could fix this:
1. If we switch to six-flavour oscillations (see #99), we wouldn’t need the L and R matrices anymore
2. If we stick with four-flavour oscillations, we could calculate the `self.De1`, `self.De2` etc. in the `__init__` function (as we already do for many of the existing transformations) and then use those in the calculation to compose flavour transformations.

jpkneller:
Either option should work. The second is the easiest and I know what to do. I haven't looked at how much of the code would have to change if we went to 6 flavors though it's something we said we'd do.

sybenzvi:
This PR is superseded by several other efforts to handle flavor transformations and is now obsolete. Closing without merging.

PR #206, opened 2022-09-07
Log Bins for snowglobes.py

stlgolfer:
This PR is to continue the discussion in the tg3-detectorresponse slack channel regarding logarithmically spaced time bins in the `generate_time_fluence` feature in `snowglobes.py`. I've also implemented @JostMigenda 's `ValueError` idea for dealing with negative times. Please note that this PR is merely a prototype and it includes two files: `data_handlers.py` and `log_bins.py` that are necessary to testing but are not intended to be included in the final product.


JostMigenda:
Thank you for this contribution!
Log-spaced time bins are a great idea, and will be especially useful for pre-SN neutrinos once we officially support them in snewpy. The code looks fine to me; after removing the testing files and commented out code (and perhaps some bits of very minor  cleanup) this should be okay to merge.

python/snewpy/snowglobes.py
```diff
@@ -96,13 +96,6 @@ def generate_time_series(model_path, model_type, transformation_type, d, output_
 
     # now process log data
     if (log_bins==True):
```
JostMigenda, line 98: Python lets you simply write `if log_bins:`, which is cleaner and easier to read.

python/snewpy/snowglobes.py
```diff
@@ -96,13 +96,6 @@ def generate_time_series(model_path, model_type, transformation_type, d, output_
...
-        needed_offset = -1*u.s
-        # need to offset the times so no negatives
-        #if tedges[0] < 0:
-            #needed_offset = tedges[0] + 0.0001*u.s
-            #tedges = tedges + tedges[0] + 0.0001*u.s #shift so it's very close to 0
-            #tmin = 0.0001*u.s
-            #tmax+=tmin
         log_edges = np.asarray([])
 
         if tmax < 0 or tmin < 0:
```
JostMigenda, line 101: Shouldn’t these be `<=0`?

python/snewpy/snowglobes.py
```diff
@@ -91,6 +94,22 @@ def generate_time_series(model_path, model_type, transformation_type, d, output_
...
+    if (log_bins==True):
+        log_edges = np.asarray([])
+
+        if tmax < 0 or tmin < 0:
+            raise ValueError("Cannot apply log to time windows that are less than 0")
+        tstep = math.log10(abs(tmax/tmin))/len(times)
+
+        for i in range(0,len(times)):
+            t = (tmin/u.s)*(10**(i*tstep))
+            if t < 0:
```
JostMigenda, line 107: Can `t` ever be less than 0 here, if we’ve already checked a few lines earlier that `tmin` is positive?

JostMigenda:
I’m so sorry, but it seems I missed the forest for the trees in my earlier review:
```
if tmax <= 0 or tmin <= 0:
    raise ValueError("Cannot apply log to time windows that are less than or equal to 0. Consider adjusting model time window")
```
The problem is that the `generate_time_series` function here doesn’t let the user select `tmin` and `tmax` (only `generate_fluence` does!), so this error message is not actionable.

These inconsistent and confusing arguments are definitely something we need to fix as part of #209 (which just got moved a little higher up my priority list). In the meantime, should we put this PR on hold? Or is it worth adding `tmin` and `tmax` arguments to `generate_time_series` in this PR as well, even though we’ll overhaul the generate functions soon anyway?

jpkneller:
We decided to close this PR - it is already implemented if the user supplies arrays for tstart and tend which are log spaced..

PR #211, opened 2022-09-19
Remove duplicate time entries from Walk_2019 nu_e_bar file

### LINKED ISSUE #210: Repeated time entries in Walk_2019 nu_e_bar file
The nu_e_bar table loaded for the Walk_2019 model contains duplicate entries for the times listed below. Column 1 is the time in s, column 2 is the index in the `Walk_2019.time` member. This was found by running the model through `scipy.interpolate.PchipInterpolator`, using `model.time` as the argument `x`.

This only features in the nu_e_bar file, and appears to have been resolved for other flavors by commit 1e37a4c

https://github.com/SNEWS2/snewpy/blob/main/models/Walk_2019/s40.0c_3DBH_dir1_LS220_nuebar

Code to replicate & find duplicate entries
```python
from snewpy.models.ccsn import Walk_2019
from scipy.interpolate import PchipInterpolator
import numpy as np

model = Walk_2019(filename='<path-to-snewpy-models>/Walk_2019/s40.0c_3DBH_dir1')
time = model.time

try:
    interp_lum = PchipInterpolator(x=time, y=model.luminosity[Flavor.NU_E_BAR])
except ValueError as e:
    print(e)

diff = np.diff(time)
idx = np.where(diff==0)[0]
print(f"{'time':<10s} {'idx':<5s}")
for t, i in zip(time[idx], idx):
    # Add +1 to index to point at duplicate entries
    print(f"{t.value:<10f} {i+1:<5d}") 
```
Expected Output
```
`x` must be strictly increasing sequence.
time       idx  
0.110000   202  
0.177700   339  
0.199450   384  
0.227050   441  
0.298840   586  
0.299840   589  
0.300840   592  
0.317840   627  
0.340850   675  
0.340850   676  
0.360000   716  
0.379800   757  
0.388780   776  
0.410860   822  
0.435360   872  
0.435360   873  
0.454740   913  
0.479720   964  
0.500000   1006 
0.521940   1051 
0.526970   1063 
0.527320   1065 
0.531420   1075 
0.560410   1134 
0.563570   1142 
0.570090   1157 
```

sgriswol:
Fixes #210


sgriswol:
Though this PR is named `fix_Walk_2019` I also found a similar issue in several of the Zha_2021 models. I have committed changes to the Zha models on [Zha_2021_duplicate_times](https://github.com/SNEWS2/snewpy/tree/Zha_2021_duplicate_times). Those model files are larger, and there are more duplicate times. To be specific, there were 3,658 duplicates out of 180,913 total entries, over 11 model files. 

Given the much broader scope of these duplicates, and that in most cases the duplication seems consistent with the time being set to a fixed precision, I think the changes to the Zha models might be worth some extra consideration. I didn't see the same pattern of fixed precision in the Walk model.

JostMigenda (approved):
LGTM, thanks!

JostMigenda:
@evanoconnor, can you weigh in regarding the Zha_2021 model?

evanoconnor:
I had a look.  Most of these do indeed look like fixed precision.  Near the end of the simulations the time step decreases to slightly below the output precision and once every 10 steps it seems to double the step.

Given the large number of points and the very small dt's here, the suggested change is perfectly fine.  

There is just one interesting case which is the s18.dat where this only occurs once, it looks like there it may have been a restart stitching issue, but again, fine to just remove the offending lines.

Issue #205, opened 2022-08-05
Error using `wc100kt30prct_he` detector


JostMigenda:
With snewpy v1.2.1 and SNOwGLoBES v1.2, when changing the detector to `wc100kt30prct_he` in the `SNOwGLoBES_usage.ipynb` example notebook, `snowglobes.simulate()` fails with a long error message:

> Error: In line 400: Value for ranges in smear out of range
> bin/supernova: ERROR in experiment supernova.glb: in smearing matrix 0, line 688: Invalid lower sampling point index: 1974755981.
> […]
> bin/supernova: ERROR in experiment supernova.glb: user_post_smearing_channel has not numofbins elements.
> bin/supernova: FATAL: Incompletely or incorrectly defined experiment!.

The `ar40kt_he` detector works, though.

PR #212, opened 2022-09-20
Fix issue with he detectors

mcolomermolla:
This should close issue #205

The problem was the following:

- guess_material() was not considering the '_he' in the material name, which was causing that efficiency and smearing files were not found, and those matrices were set to ones. This was producing no error in the ar40kt_he case, but it was not properly running with he eff and smear files.
- for the case of wc100kt30prct_he detector, the oxygen files don't contain '_he' in the channel, but ibd and nu-e ES do. So for the oxygen case, the file name was correctly found, but there was a missmatch with the rates binning, which was set only for the non he case.

### DIFF
--- a/python/snewpy/snowglobes_interface.py
+++ b/python/snewpy/snowglobes_interface.py
@@ -160,7 +160,7 @@ class SimpleRate():
         data = {}
         energies = np.linspace(7.49e-4, 9.975e-2, 200) # Use the same energy grid as SNOwGLoBES
         if '_he' in detector:
-            energies = np.linspace(7.49e-4, 2*9.975e-2, 400) #SNOwGLoBES grid for he configurations
+            energies = np.linspace(7.49e-4, 19.975e-2, 400) #SNOwGLoBES grid for he configurations
         for channel in self.channels[material].itertuples():
             xsec_path = f"xscns/xs_{channel.name}.dat"
             xsec = np.loadtxt(self.base_dir/xsec_path)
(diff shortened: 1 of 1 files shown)


python/snewpy/snowglobes_interface.py
```diff
@@ -156,6 +159,8 @@ def _compute_rates(self, detector, material, flux_file:Path):
         TargetMass = self.detectors[detector].tgt_mass
         data = {}
         energies = np.linspace(7.49e-4, 9.975e-2, 200) # Use the same energy grid as SNOwGLoBES
+        if '_he' in detector:
+            energies = np.linspace(7.49e-4, 2*9.975e-2, 400) #SNOwGLoBES grid for he configurations
```
JostMigenda, line 163: Should the upper edge really be 2*99.75 = 199.5 MeV here? Or should it be 199.75 MeV? (I’d always assumed that the reason for the 0.75 was simply that this is the midpoint of a 0.5 MeV bin; but I can’t see this actually documented anywhere.)

JostMigenda (commented):
Other than this one binning question, this looks good to me! Thank you very much for the explanation!

(And indeed, I have verified that the previous snewpy version used efficiency/smearing factors of 1 for `ar40kt_he`, so unsmeared and smeared event numbers were the same as you explained.)

python/snewpy/snowglobes_interface.py
mcolomermolla: Yes, you're right, it should indeed be the bin center, with 0.5MeV bins (snowglobes choice), I will modify that (since the bins are twice, from 200 to 400, I assumed a x2 for the upper edge was correct). Indeed, there is no documentation on this. SimpleRate just kept here the choice of the binning by snowglobes, and when moving from root histograms to python arrays, kept the bin center to define the points in the list.

JostMigenda (approved):
Thanks for the fix!

PR #217, opened 2022-10-06
Moving reading flux files to SimpleRate.run function, remove hardcoded binning

### LINKED ISSUE #216: snowglobes.SimpleRate: add method for rate calculation from the given flux array
# What I want

To be able to calculate the rate from the neutrino flux stored as `np.array` in memory:
```python
sng = SimpleRate()
result = sng.compute_rates(flux_array, energies, detector, material)
```

# What we have

There are only methods for calculating rates reading the flux files from the disk:
https://github.com/SNEWS2/snewpy/blob/b35379952fe18aedc28cdb106f1ee9e7ad08b082/python/snewpy/snowglobes_interface.py#L156
https://github.com/SNEWS2/snewpy/blob/b35379952fe18aedc28cdb106f1ee9e7ad08b082/python/snewpy/snowglobes_interface.py#L205

# What I suggest

A small refactoring inside `SimpleRate` class:
1. Transform 
```python 
_compute_rates(self, detector, material, flux_file:Path)
``` 
to
```python
compute_rates(self, flux:np.ndarray, energies:np.ndarray, detector, material)
#Note: also making the function public
```
and move reading the arrays from files to `SimpleRate.run`

### LINKED ISSUE #218: snowglobes.SimpleRate - remove hardcoded binning
# What I want

As the developer I want our `snowglobes_interface` to be easily adaptable for the new materials in SNOwGLoBES. New materials can have different energy binning, so I'd like `SimpleRate` to be able to read the binning from SNOwGLoBES and use it when computing the rate

# What we have now
I see that the energy binning in SimpleRate is defined as
https://github.com/SNEWS2/snewpy/blob/b35379952fe18aedc28cdb106f1ee9e7ad08b082/python/snewpy/snowglobes_interface.py#L161-L163

# What I suggest

We are already reading this binning for each material:
https://github.com/SNEWS2/snewpy/blob/b35379952fe18aedc28cdb106f1ee9e7ad08b082/python/snewpy/snowglobes_interface.py#L101-L107
so there is no need to guess energies from the material name - we could use `self.binning[material]` - after we transform it to the binning array(s).

Sheshuk:
Work on `SimpleRate`
Closes #216 
Closes #218

### DIFF
--- a/python/snewpy/snowglobes_interface.py
+++ b/python/snewpy/snowglobes_interface.py
@@ -198,7 +198,7 @@ class SimpleRate():
             flavor = flavor_index + (3 if channel.parity == '-' else 0)
             flux = fluxes[flavor]
             # Cross-section in 10^-38 cm^2
-            xsecs = np.interp(np.log(energies)/np.log(10), xsec[:, 0], xsec[:, 1+flavor], left=0, right=0) * energies_t
+            xsecs = np.interp(np.log(energies_t)/np.log(10), xsec[:, 0], xsec[:, 1+flavor], left=0, right=0) * energies_t
             # Fluence (flux integrated over time bin) in cm^-2 
             # (must be divided by 0.2 MeV to compensate the multiplication in generate_time_series)
             fluxs = np.interp(energies_t, flux_energies, flux, left=0, right=0)/2e-4
@@ -211,8 +211,8 @@ class SimpleRate():
             data[(channel.name,'unsmeared','weighted')] = weighted_rates
             # Add detector effects
             if self.smearings and self.efficiencies:
-                smear = self.smearings[detector].get(channel.name, default=np.eye(len(rates)))
-                effic = self.efficiencies[detector].get(channel.name, default=np.ones(len(rates)))
+                smear = self.smearings[detector].get(channel.name, np.eye(len(rates)))
+                effic = self.efficiencies[detector].get(channel.name, np.ones(len(rates)))
                 rates = np.dot(smear,rates) * effic
                 weighted_rates = rates * channel.weight
                 # Write to dictionary
--- a/python/snewpy/snowglobes_interface.py
+++ b/python/snewpy/snowglobes_interface.py
@@ -104,7 +104,10 @@ class SimpleRate():
                 if not l.startswith('%'):
                     l = '% 200 0.0005 0.100 200 0.0005 0.100'
                 tokens = l.split(' ')[1:]
-                return dict(zip(['nsamples','smin','smax','nbins','emin','emax'],tokens))
+                nsamples,smin,smax, nbins,emin,emax = [float(t) for t in tokens]
+                return {'e_true' :np.linspace(smin,smax,nsamples),
+                        'e_smear':np.linspace(emin,emax,nbins)
+                        }
 
         self.channels = {}
         self.binning = {}
--- a/python/snewpy/snowglobes_interface.py
+++ b/python/snewpy/snowglobes_interface.py
@@ -154,6 +154,30 @@ class SimpleRate():
         logger.debug(f'smearing matrices: {self.smearings}')
 
     def compute_rates(self, detector:str, material:str, fluxes:np.ndarray, energies:np.ndarray):
+        """ Calculate the rates for the given neutrino fluxes interacting in the given detector.
+
+        Parameters
+        ----------
+        detector: str
+            Detector name in SNOwGLoBES. Check `SimpleRate.detectors` for the list of options
+        material: str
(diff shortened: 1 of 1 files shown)


python/snewpy/snowglobes_interface.py
```diff
@@ -153,26 +156,52 @@ def _load_smearing_matrices(self, path):
...
-        if '_he' in detector:
-            energies = np.linspace(7.49e-4, 19.975e-2, 400) #SNOwGLoBES grid for he configurations
+
+        #load the binning for the smearing
+        binning= self.binning[material]
+        #calculate bin centers 
+        energies_t = 0.5*(binning['e_true'][1:]+ binning['e_true'][:-1] )
+        energies_s= 0.5*(binning['e_smear'][1:]+binning['e_smear'][:-1])
+        binsize = np.diff(binning['e_true'])
+
```
mcolomermolla, line 193: In principle, truth and smeared energies should have the same binning, i.e. be the same (just # of events per energy bin will be different). In fact, if their binning is not equal, there should be an error. So, not sure we want to make the distinction.

python/snewpy/snowglobes_interface.py
```diff
@@ -104,7 +104,10 @@ def _read_binning(fname):
                 if not l.startswith('%'):
                     l = '% 200 0.0005 0.100 200 0.0005 0.100'
                 tokens = l.split(' ')[1:]
-                return dict(zip(['nsamples','smin','smax','nbins','emin','emax'],tokens))
+                nsamples, smin,smax, nbins,emin,emax = [float(t) for t in tokens]
+                return {'e_true' :np.linspace(smin,smax,int(nsamples)+1),
+                        'e_smear':np.linspace(emin,emax,int(nbins)+1)
+                        }
```
mcolomermolla, line 110: agreed with this removal of hardcoded binning

python/snewpy/snowglobes_interface.py
```diff
@@ -239,11 +261,14 @@ def run(self, flux_files, detector:str, material:str=None):
...
-        
+
         with tqdm(total=len(flux_files), leave=False, desc='Flux files') as progressbar:
             results = []
             for flux_file in flux_files:
-                result = self._compute_rates(detector,material,Path(flux_file))
+                e_flux = np.loadtxt(Path(flux_file).resolve()).T
+                energies = e_flux[0]
+                fluxes = e_flux[1:]
+                result = self.compute_rates(detector, material, fluxes, energies)
```
mcolomermolla, line 271: I also agree with this change as discussed in the issue, better fluxes and energies than the path, and removing the underscore

python/snewpy/snowglobes_interface.py
Sheshuk: I just followed the format from (SNOw)GLoBES.

1. GLoBES uses distinct values for "energy" (e_smear) and "sample" (e_true) values, and that is what it expected for the smearing matrices, as I understood.
2. The code, generating the smearing matrices in SNOwGLoBES has these binning parameters separated:
https://github.com/SNOwGLoBES/snowglobes/blob/958cd2d891c873cd7e89addc1870d9bc940109f0/smearing_code/create_smearing_matrix.py#L24-L47

So I decided that we should conform to that, even if now `e_true` and `e_smear` are the same in all the smearing matrices, they might be different in the future. 

> In principle, truth and smeared energies should have the same binning

I think for some cases it would make sense them to be different. If `e_smear` is the detected energy, then for, say, neutron capture channels (like in HALO) we don't see the energy spectrum at all, just some monoenergetic lines. Then it might make sense to have smearing matrix for these channels to translate from full `e_true` range to several `e_smear` bins.
So I don't think we should raise an error if they are different - the code should work correctly for those.

mcolomermolla: OK, indeed the matrix does not have to be squared (even if that's the case for the detectors implemented in snowglobes now) but to go from e_truth to e_smeared, the smearing_matrix needs to have same true energy binning as the e_truth to match the input for the multiplication, that's what I meant. 

I think you can keep the distinction as it is with no problem.

Sheshuk: Thanks! 

> smearing_matrix needs to have same true energy binning as the e_truth

Yes, I absolutely agree with that. As I understand the binning of the smearing matrix is what is written out in the `channels/channels_{material}` for some of the materials - and that's what we're reading in `_read_binning`.

If you agree on this implementation, could you please approve this PR?

mcolomermolla: Yes!

mcolomermolla (approved):

PR #220, opened 2022-10-11
Adding get_flux function to SupernovaModel

### LINKED ISSUE #213: Add "calculate neutrino flux at given distance" to SupernovaModel
## What I want

As a user I would like to get the supernova neutrino flux, predicted by a certain `model` (subclass of `SupernovaModel`) at a given `distance` at given times and energies (in `neutrinos/cm^2/MeV/s`).

## What we have now
Currently `SupernovaModel` class allows only to get neutrino flux in 4pi (`get_initial_spectra`, `get_transformed_spectra`), and has no means for calculating flux at a given distance.

We have these calculations only in `generate_fluence` and `generate_time_series`:
https://github.com/SNEWS2/snewpy/blob/b35379952fe18aedc28cdb106f1ee9e7ad08b082/python/snewpy/snowglobes.py#L122
https://github.com/SNEWS2/snewpy/blob/b35379952fe18aedc28cdb106f1ee9e7ad08b082/python/snewpy/snowglobes.py#L288

which will finally save the flux/fluence to the disk, which is not always needed.

While these calculations are not complicated, we shouldn't make user I think they belong to the SupernovaModel class.

## What I suggest

1. Create a method in the base class `SupernovaModel.get_initial_flux (t, E, distance, **kwargs)`
which by default would call `self.get_initial_spectra`. If there are any additional input parameters for a specific model (direction? flavors?), this method should be reimplemented in the models's class, making use of `**kwargs` argument.

2. This method should return an object, containing the flux for each of the `(Flavor,T,E)` points, and optionally pack

    * Energy points `E`
    * Time points `T`
    * Available flavors 

   In a simplest case it could be `Dict[Flavor, np.array]`, but I propose to discuss this separately in #214

Sheshuk:
Closes #213

### DIFF
--- a/python/snewpy/models/base.py
+++ b/python/snewpy/models/base.py
@@ -162,7 +162,7 @@ class SupernovaModel(ABC):
 
         return transformed_spectra   
 
-    def get_flux (self, t, E, distance, flavor_xform=NoTransformation, **kwargs):
+    def get_flux (self, t, E, distance, flavor_xform=NoTransformation(), **kwargs):
         """Get neutrino flux through 1cm^2 surface at the given distance
 
         Parameters
@@ -185,7 +185,7 @@ class SupernovaModel(ABC):
         """
         distance = distance << u.kpc #assume that provided distance is in kpc, or convert
         factor = 1/(4*np.pi*(distance.to('cm'))**2)
-        flux = get_transformed_spectra(self, t, E, flavor_transform)
+        flux = self.get_transformed_spectra(t, E, flavor_xform)
         return {flavor: f*factor for flavor,f in flux.items()}
 
 
--- a/python/snewpy/models/base.py
+++ b/python/snewpy/models/base.py
@@ -9,6 +9,7 @@ from astropy.units.quantity import Quantity
 from scipy.special import loggamma
 
 from snewpy.neutrino import Flavor
+from snewpy.flavor_transformation import NoTransformation, FlavorTransformation
 from functools import wraps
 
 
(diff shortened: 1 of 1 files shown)


python/snewpy/models/base.py
```diff
@@ -161,6 +162,33 @@ def get_transformed_spectra(self, t, E, flavor_xform):
 
         return transformed_spectra   
 
+    def get_flux (self, t, E, distance, flavor_xform=NoTransformation(), **kwargs):
```
JostMigenda, line 165:
```suggestion
    def get_flux (self, t, E, distance, flavor_xform=NoTransformation()):
```
If the `**kwargs` aren’t used (or even passed through to some other function), I would delete them here.

python/snewpy/models/base.py
```diff
@@ -9,6 +9,7 @@
 from scipy.special import loggamma
 
 from snewpy.neutrino import Flavor
+from snewpy.flavor_transformation import NoTransformation, FlavorTransformation
```
JostMigenda, line 12:
```suggestion
from snewpy.flavor_transformation import NoTransformation
```
I don’t think `FlavorTransformation` is used anywhere in this PR?

python/snewpy/models/base.py
```diff
@@ -161,6 +162,33 @@ def get_transformed_spectra(self, t, E, flavor_xform):
...
+        """Get neutrino flux through 1cm^2 surface at the given distance
+
+        Parameters
+        ----------
+        t : astropy.Quantity
+            Time to evaluate the neutrino spectra.
+        E : astropy.Quantity or ndarray of astropy.Quantity
+            Energies to evaluate the the neutrino spectra.
+        distance : astropy.Quantity or float (in kpc)
+            Distance from supernova to the
```
JostMigenda, line 175:
```suggestion
            Distance from supernova.
```

python/snewpy/models/base.py
```diff
@@ -161,6 +162,33 @@ def get_transformed_spectra(self, t, E, flavor_xform):
...
+            Energies to evaluate the the neutrino spectra.
+        distance : astropy.Quantity or float (in kpc)
+            Distance from supernova to the 
+        flavor_xform : FlavorTransformation
+            An instance from the flavor_transformation module.
+
+        Returns
+        -------
+        dict
+            Dictionary of neutrino fluxes in [neutrinos/(cm^2*MeV*s)],
```
JostMigenda, line 182:
```suggestion
            Dictionary of neutrino fluxes in [neutrinos/(cm^2*erg*s)], 
```
`get_transformed_spectra` returns `<Quantity 1.234e+56 1 / (erg s)>`

JostMigenda (commented):
Looks good to me, just some very minor comments

python/snewpy/models/base.py
Sheshuk: You're right. What I had in mind is to be compatible with all the models, including Fornax etc. But this is a separate issue #223 
For this PR I should remove this.

JostMigenda (approved):

PR #222, opened 2022-10-11
Using time arrays in PinchedModel.get_initial_spectra

### LINKED ISSUE #221: Allow array as `t` argument for `SupernovaModel.get_*_spectra`
# What I want

As a user I need to build the neutrino spectrum vs. time. I want to be able to calculate this 2d spectra directly from `SupernovaModel` without constructing loops on the `t` variable:

```python
#Loading the SN model
model = ccsn.Bollig_2016('models/Bollig_2016/s11.2c')
#time and energy bins
t = np.linspace(0,10,101)<<u.s
E = np.linspace(0,50,51)<<u.MeV

fluxes2d = model.get_initial_spectra(t,E)
assert fluxes2d[0].shape == (101, 51)

#backward compatibility: if we provide a scalar time we get 1D arrays.
#Let's take t[20] as scalar value
fluxes1d = model.get_initial_spectra(t[20],E)
assert fluxes1d[0].shape == (51)

#check that 1d and 2d results are compatible
for flavor in fluxes2d:
   assert np.allaclose(fluxes2d[flavor][20], fluxes1d)
```
# What we have now
Currently `SupernovaModel.get_initial_spectra` and `SupernovaModel.get_transformed_spectra` can work only with a single `t` value. As far as I understand, the only limitation is in the implementration of `PinchedModel`, so it can be easily corrected.

This is in addition to the existing `test_snowglobes.py` tests, which should work without modification and provide crosscheck with the events table.
# Outlook
This should make possible further optimization in `snowglobes.generate_*` functions - getting rid of the time loops there can speed up computation and make code cleaner.

After implementing #214 we can use `NeutrinoFlux` class as the return value.

Sheshuk:
Closes #221

### DIFF
--- a/python/snewpy/models/base.py
+++ b/python/snewpy/models/base.py
@@ -238,6 +238,9 @@ class PinchedModel(SupernovaModel):
         initialspectra : dict
             Dictionary of model spectra, keyed by neutrino flavor.
         """
+        t = u.Quantity(t, ndmin=1)
+        E = u.Quantity(E, ndmin=2)
+
         initialspectra = {}
 
         # Avoid division by zero in energy PDF below.
@@ -258,14 +261,18 @@ class PinchedModel(SupernovaModel):
             a  = np.interp(t, self.time, self.pinch[flavor])
 
             # Sanity check to avoid invalid values of Ea, alpha, and L.
-            initialspectra[flavor] = np.zeros_like(E, dtype=float) / (u.erg*u.s)
-            if L <= 0. or Ea <= 0. or a <= -2.:
-                continue
+            L  = np.expand_dims(L, axis=1)
+            Ea = np.expand_dims(Ea,axis=1)
+            a  = np.expand_dims(a, axis=1)
             # For numerical stability, evaluate log PDF and then exponentiate.
-            initialspectra[flavor] = \
+            result = \
               np.exp(np.log(L) - (2+a)*np.log(Ea) + (1+a)*np.log(1+a)
                     - loggamma(1+a) + a*np.log(E) - (1+a)*(E/Ea)) / (u.erg * u.s)
-
+            #remove bad values
+            result[np.isnan(result)] = 0
+            #remove unnecessary dimensions, if E or t was scalar:
+            result = np.squeeze(result)
+            initialspectra[flavor] = result
         return initialspectra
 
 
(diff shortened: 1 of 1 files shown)


python/snewpy/models/base.py
```diff
@@ -238,6 +238,9 @@ def get_initial_spectra(self, t, E, flavors=Flavor):
         initialspectra : dict
             Dictionary of model spectra, keyed by neutrino flavor.
         """
+        t = u.Quantity(t, ndmin=1)
+        E = u.Quantity(E, ndmin=2)
+
```
mcolomermolla, line 243: I don't understand why both arrays don't have ndmin = 1 but for energy it is 2.

Sheshuk: You're right, it's not very clear here. 
1) I wanted to ensure that both `t` and `E` are converted to arrays even if they are only scalars
2) I wanted them to be 2D arrays, so everything relevant to `E` would be of shape `(len(E),1)` and `t` would be `(1, len(t))` - so after the calculations we'll have 2D array of shape `(len(E), len(t))`. **But** `t` will be used later in the `np.interp` method, which works only for 1D arrays. So I had to keep it 1D, and produce 1D arrays `L`,`Ea`,`a`, and later turn them to shape `[1, len(t)]` using `np.expand_dims`.

I will update the code to make it more clear.

Sheshuk: Please have a look now

mcolomermolla: Okey, thanks for clarifying, I get now what you did, it makes sense to me

JostMigenda:
Minor note: If `t` can now also be an array, the type of `t` in the docstring should be updated accordingly.

Sheshuk:
> Minor note: If `t` can now also be an array, the type of `t` in the docstring should be updated accordingly.

`astropy.Quantity` is an extension of ndarray, so we're actually using them, not `ndarray of astropy.Quantity`, for example
```python
>>> T = np.linspace(0,10) << u.s
>>> type(T)
astropy.units.quantity.Quantity
```
so the current docstring is (now even more) correct

JostMigenda (approved):
Good point about the types! It’s a bit unfortunate that this distinction cannot be expressed right now, but maybe that’ll change in the coming years as a result of [PEP 646](https://peps.python.org/pep-0646/).

This PR LGTM; thanks, Andrey!

Sheshuk:
Thanks! Yes these variadic generics look like they would make the signatures much cleaner.
In general I think it would be good to add the type hints throughout snewpy, but that's a separate task.

PR #226, opened 2022-10-13
Fix integration test

sgriswol:
Fixes failing integration test, the result of a call between @JostMigenda and myself. The error was related to handling of absolute paths to model files.

### DIFF
--- a/python/snewpy/_model_downloader.py
+++ b/python/snewpy/_model_downloader.py
@@ -100,8 +100,7 @@ local path.
             logger.info(f'Downloading file {self.path}')
             _download(self.remote, self.path)
             self.check()
-        finally:
-            return self.path
+        return self.path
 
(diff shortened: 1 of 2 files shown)


python/snewpy/_model_downloader.py
```diff
@@ -100,8 +100,7 @@ def load(self) -> Path:
             logger.info(f'Downloading file {self.path}')
             _download(self.remote, self.path)
             self.check()
-        finally:
-            return self.path
```
JostMigenda, line 104: Just a note: This change is not strictly necessary to fix the integration test, but it was something we discovered while debugging—if the return statement is in a `finally` block, any errors occuring in the `except` block (e.g. during the download) are suppressed.

JostMigenda (approved):
LGTM

PR #128, opened 2021-11-04
Prevent running integration tests on draft PR

Sheshuk:
I think my commits might have been heavy for the runner, sorry.
Let's run the tests only when pull request is in a ready state.

PR #227, opened 2022-10-13
Skip integration test for draft PRs

JostMigenda:
In #128, we changed the GitHub workflow so that integration test would not run on draft PRs. Unfortunately, with the solution we found back then, it required manual activity by a developer to run; so it did not run reliably on PRs, which has caused some problems recently.

After looking at the GitHub documentation some more, I think I’ve found a better way to achieve the same thing. I will test in this PR.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -9,6 +9,7 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+    types: [opened, synchronize, reopened, ready_for_review]
 
 jobs:
   run:
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -9,10 +9,12 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
-    types: [review_requested, ready_for_review]
 
 jobs:
   build:
+    # Only execute integration test if PR is *not* a draft
+    if: !github.event.pull_request.draft
+
     runs-on: ubuntu-latest
     strategy:
       # Add a list of python versions we want to use for testing.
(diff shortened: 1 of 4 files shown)


JostMigenda:
In [728da55](https://github.com/SNEWS2/snewpy/pull/227/commits/728da555e9ff512dfac3ef796f66c7f893dd9833) you can see nicely that the “Integration Tests / build” job is skipped for this draft PR.

JostMigenda:
Only downside now is that the integration test doesn’t re-run if a draft is marked as ready for review. But that should be easy to fix in the following commit

(FWIW, GitHub documentation for the `pull_request` event is [here (for activity types)](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) and [here (for the format of `github.event.pull_request`)](https://docs.github.com/en/rest/pulls/pulls#get-a-pull-request).)

JostMigenda:
Before we review and merge this, one more question: As part of the integration test, we currently run a timing benchmark and publish the results in a comment on the PR. (@Sheshuk originally added that in https://github.com/SNEWS2/snewpy/pull/132/commits/3bb7899c7a01b524d0b1fe499569b8c16d5820c8)

Over the past year, I don’t remember us ever using the results of that benchmark, because they are so unreliable—it’s random chance whether the workflow runs on a GitHub worker machine that happens to be fast or slow. (E.g. in the comment above, if you view the edit history you can see that the mean time was between 3.49 and 5.47 for the exact same commit.) Instead, we run manual benchmarks anyway whenever we discuss performance improvements.
I would therefore suggest we remove this benchmark (and the workflow that publishes the results).

mcolomermolla (commented):
I agree with removing benchmark timing from integration test, it is indeed completely randomly depending on the machine. Then, it can be merged.

JostMigenda:
Okay, I’ve removed all the benchmarking code. @mcolomermolla, do you want to take a quick look and post an approving review if you’re happy with this?

mcolomermolla (approved):

PR #229, opened 2022-11-07
Basic support for SNOwGLoBES v1.3

### LINKED ISSUE #201: SNOwGLoBES 1.3 Compatibility
Currently the channel file structure for custom binning breaks SNEWPY's channel.dat dataframe. This is due to the 'SN_nu' tag put in the beginning of each line (needed to use a the custom binning postamble file ). 

To get rid of the SN_nu tag while making the df use the following code 
```
df_chan = pd.read_csv(path_chan, sep=" ", header=None)

if df_chan[0][0] == "SN_nu":
    df_chan.drop(df_chan.columns[0], axis=1, inplace=True)
    
# then you can add the col names.. 
```

JostMigenda:
Fixes #201 and should be used together with the upcoming version of SNOwGLoBES (see PR SNOwGLoBES/snowglobes#19).

Note: To support CEνNS channels, we would need to allow channel-specific energy binning; whereas the `SimpleRate` implementation currently assumes that all channels for a given material use the same energy binning.
This PR does not yet support CEνNS channels; it only includes changes to ensure that existing code keeps working with the new SNOwGLoBES version.


JostMigenda:
Note: The automated integration tests currently fail because the smearing matrices are in a different directory in SNOwGLoBES 1.3; so when testing this PR with SNOwGLoBES 1.2, it doesn’t find the smearing matrices and fails the test.

I’ve tested it locally, but will be sure to merge the SNOwGLoBES PR first and then re-run the tests before merging this PR.

mcolomermolla (approved):
I have tested it locally with full installation of the new version and there are no problems. We can move forward with merging.

JostMigenda:
Hmm. On rerunning the integration tests, there are some more issues related to the `icecube` and `ar40kt` smearing files provided by SNOwGLoBES. The IceCube one is straightforward, I'll fix that in the next few days. The other one may take longer.

JostMigenda:
I just re-ran the tests after @Storreslara and I made the necessary SNOwGLoBES fixes. Everything’s working now; so we can
- [ ] tag SNOwGLoBES v1.3
- [ ] update snewpy docs and workflows to refer to SNOwGLoBES v1.3
- [ ] …and then merge this PR.

I feel like I’m about to come down with a flu; but if I’m well enough, I can try to do this after the SNEWS call tomorrow.

Issue #225, opened 2022-10-13
`init_model` convenience function


JostMigenda:
The `init_model` convenience function was added in #184 to try out; but is currently hidden in `snewpy.models.util`, which is (a) not included in the docs and (b) probably not a submodule that we’d want to expose to the user.

There’s also [Andrey’s comment](https://github.com/SNEWS2/snewpy/pull/184#discussion_r824550344) about whether to require e.g. `"ccsn.Nakazato_2013"` as an argument instead of just `"Nakazato_2013"`.

JostMigenda:
I’ve [moved the function](https://github.com/SNEWS2/snewpy/pull/230/commits/0c86d793d572d3a7e8a953a13c058cb7c59ef6e9) to `snewpy.models` in #230.

JostMigenda:
Also: Should this still use the `get_models` downloader (which downloads all progenitors for that model family) or should it use the new downloader (which only loads the requested progenitor)?

JostMigenda:
Almost two years later, we’ve never actually needed this convenience function; so let’s remove this dead code.

PR #230, opened 2022-11-11
Model registry cleanup

### LINKED ISSUE #224: Cleanup of model registry
Two points that were left over from #184 but are important to fix before a final release:
* [ ] update Jupyter notebooks and “Getting Started” page of documentation with new way of initialising models
* [ ] Clean up the models in `ccsn`, `base` and `loader`. (E.g. unused `_RegistryModel` base class, duplicate code in loader classes)

As part of cleanup, see also [this comment](https://github.com/SNEWS2/snewpy/pull/184/files/d1dfec61831006f5dc9143043a84ba552cb1f27f#r824553338) by @Sheshuk.

JostMigenda:
This fixes #224. Two things I want to highlight in particular:
* It implements [this suggestion](https://github.com/SNEWS2/snewpy/pull/184/files/d1dfec61831006f5dc9143043a84ba552cb1f27f#r824553338) by @Sheshuk to move the parameter checking into the appropriate base class
* It removes the `Model.param_combinations` property in favour of `Model.get_param_combinations()` (which already existed anyway). In addition to saving some boilerplate code, this is also more extensible. For example, I could imagine adding the ability to give additional arguments like `get_param_combinations(eos='shen')` to narrow the list down.

During the review, please note that thus far, I’ve only updated `Nakazato_2013.ipynb` with the new model initialisation. Let me know if that looks reasonable; if so, I will update
- [x] the notebooks for all other models and
- [x] the [Getting Started](https://snewpy.readthedocs.io/en/stable/gettingstarted.html) page of the documentation

before we can merge this PR.


sgriswol (changes requested):
This looks great! The comments I have should be very easy to address. The most important one being, I think `snewpy.models._init_model` should be public-facing. Is there a reason it is named with a leading underscore?

JostMigenda:
> `snewpy.models._init_model` should be public-facing.

Eventually, yes. But for now, I think it’s not yet ready—see #225. For this PR, I just wanted to move that out of the `util.py` file so I could delete that as part of the cleanup.

sgriswol (changes requested):
Thanks for putting this all together @JostMigenda! I have a few comments, most of which are cosmetic and should be easy to address. 
- [ ] The citation links `OConnor_2013.ipynb` , `Zha_2021.ipynb`, and `Warren_2020.ipynb` are missing.
- [ ] The output of code cell  3 in `Nakazato_2013.ipynb` is rather hard to read. 
- [ ] The population of valid `eos` and `progenitor_mass` parameters in the docstring of `snewpy.models.ccsn.OConnor_2013` isn't happening because [python/snewpy/models/ccsn.py:415](https://github.com/SNEWS2/snewpy/blob/dd4b05593aadadcce12f60cd1f3bbb72ef981adb/python/snewpy/models/ccsn.py#L415) is commented.
- [ ] In many of the example notebooks I received a warning over importing from `snewpy.models` rather than `snewpy.models.ccsn`. I have added suggestions to change most notebooks, but I was unable to make a suggestion for `Fornax_2019.ipynb` or `OConnor_2013.ipynb`

I will push a commit shortly to address these comments, aside from the suggestions I was able to make as part of the review (apologies for the mixed formats!).

sgriswol (approved):
Looks good to me!

Issue #199, opened 2022-08-02
Use same requirements in all SNEWS tools: SNEWPY, snewps-pt, snews-cs, snewpdag


mcolomermolla:
As an example, now:
snews-pt 1.0.2.dev4 requires pandas~=1.4.2, while SNEWPY requires pandas 1.3.4, which is incompatible.

JostMigenda:
@Sheshuk, you added the `pandas==1.3.4` requirement in https://github.com/SNEWS2/snewpy/commit/d126da84a4fb1d79ee1ff6da0f124feb32c5ece2#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552—was there any reason for this strict requirement or can we relax that?

Sheshuk:
Sorry, missed this question. There is no reason for strict requirement, we can replace = to >=1.3.4 or just remove the version. 1.3.4 was the version I used at that moment, in principle there are no restrictions.

PR #234, opened 2022-12-02
add Python 3.11 to test suite; drop 3.7

JostMigenda:
Python 3.11 [was released in late October](https://www.python.org/downloads/release/python-3110/), so this PR adds it to our testing matrix. Looking at the [3.11 release notes](https://docs.python.org/3.11/whatsnew/3.11.html), we should need no further changes for compatibility. Unfortunately, one of snewpy’s dependencies (`h5py`) [hasn’t yet published](https://github.com/h5py/h5py/pull/2165#issuecomment-1302303865) a release that supports 3.11, so unless we install `h5py` from source, snewpy will not yet install under 3.11; so this PR is currently a draft.

At the same time, I would suggest dropping support for Python 3.7 with the next snewpy release. This is in line with much of the scientific Python ecosystem (who [dropped 3.7 support in December 2021](https://numpy.org/neps/nep-0029-deprecation_policy.html)); official end of life for Python 3.7 is [June 2023](https://peps.python.org/pep-0537/#lifespan).

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       # Add a list of python versions we want to use for testing.
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.8', '3.9', '3.10', '3.11']
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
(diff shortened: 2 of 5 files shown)


JostMigenda:
One more thing on dependencies and Python 3.11: Requiring `pandas == 1.3.4` (which doesn’t have wheels for Python 3.11 available, since it was released much earlier) means we’d have to build `pandas` from source to run the workflow under 3.11. I’ll remove that requirement, so we don’t waste resources; this also fixes #199.

JostMigenda:
New h5py release with binary wheels for Python 3.11 is out now, so this PR is ready for review.

mcolomermolla (approved):
No changes requested, PR approved.

PR #235, opened 2023-02-21
Use snowglobes_data package instead of hardcoded path

### LINKED ISSUE #208: Bundle SNOwGLoBES data files with snewpy releases?
As @sybenzvi suggested in #26, it would be nice to bundle SNOwGLoBES data files (detector definitions, smearings, efficiencies, cross sections) with snewpy, so it can be installed with a single `pip install snewpy` rather than requiring a separate download.

Tying snewpy versions to specific SNOwGLoBES data file versions has both benefits and drawbacks:
On the plus side, it would make snewpy outputs more reproducible (because a given snewpy release would always use known smearing files) and it would eliminate any compatibility issues (as in #201) that users might encounter when a new SNOwGLoBES version is released that isn’t supported by snewpy yet.
The down side is that if a backwards-compatible SNOwGLoBES release adds e.g. a new detector, we would need to release a new snewpy version to update bundled files. (However, given that SNOwGLoBES updates don’t happen very often and making a new snewpy release only takes a few minutes, I don’t think this is a serious problem.)

Before implementing that, we should first ask SNOwGLoBES maintainers if they’re happy with that and check the SNOwGLoBES license for legal requirements (include license file, copyright notice, etc.)

JostMigenda:
Closes #208. I’ve just filed SNOwGLoBES/snowglobes#21, which would help publish SNOwGLoBES data files as a PyPI package. This would simplify snewpy installation (users no longer need to download SNOwGLoBES manually) and usage (no need to specify the SNOwGLoBES directory in `snewpy.snowglobes` function calls).

To test this:
1. Run `pip install -i https://test.pypi.org/simple/ snowglobes_data`
2. Then install snewpy from this branch
3. Open `SNOwGLoBES_usage.ipynb` and set `SNOwGLoBES_path` to `None`.

This PR is currently marked as a draft, because there are a few open To-Dos:
- [x] Update documentation
- [x] Update example notebook
- [x] Remove need for backported `importlib_resources` library under Python 3.9
- [x] Wait until SNOwGLoBES PR is merged & package is available on regular PyPI.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -36,16 +36,7 @@ jobs:
       - name: Install SNEWPY
         run: |
           python setup.py install --user
-      - name: Install SNOwGLoBES
-        run: |
-          mkdir opt
-          cd opt
-          git clone https://github.com/SNOwGLoBES/snowglobes.git
-          cd snowglobes
-          git checkout v1.3
-          export SNOWGLOBES=${PWD}
       - name: Run Integration Tests
         run: |
-          export SNOWGLOBES=${GITHUB_WORKSPACE}/opt/snowglobes
           python -m unittest python/snewpy/test/simplerate_integrationtest.py
           pytest -m 'snowglobes'
--- a/README.md
+++ b/README.md
@@ -31,17 +31,6 @@ To contribute, first clone the repository (`git clone https://github.com/SNEWS2/
 Once you’re happy with your changes, please submit a pull request.
 Unit tests will run automatically for every pull request or you can run them locally using `python -m unittest python/snewpy/test/test_*.py`.
 
-### Dependencies 
-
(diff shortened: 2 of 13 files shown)


Sheshuk:
I tested and it works for me. Also I think that's a really nice improvement for the SNEWPY usability - it will allow `pip install`ing it with all the dependencies rightaway 👍🏻

JostMigenda:
The SNOwGLoBES PR is now merged and [`snowglobes_data` is published on PyPI](https://pypi.org/project/snowglobes-data/). With that, I think this PR is ready.

@mcolomermolla Can you please confirm whether I’ve resolved all your comments?

mcolomermolla (approved):
Everything is fine now! Ready to be merged

PR #236, opened 2023-05-13
Implementation of NeutrinoFlux container

### LINKED ISSUE #214: Create a `NeutrinoFlux` class as a container for neutrino signal, produced by a supernova model
This is a discussion for a feature if we agree on #213.

## What I want
A `NeutrinoFlux` class, returned by `SupernovaModel.get_initial_flux` method.

An object which contains all the information about the neutrino signal from supernova model.
This information should be sufficient to 
* _Integrate the flux over time and/or energy_ - so it should have `T` and `E` points where it was calculated 
* _Apply an arbitrary neutrino mixing_ - so it needs to contain flux for all neutrino flavors

## What I suggest
1. I suggest using the `astropy.Quantity` classes to keep track of all the dimensions.
2. I suggest to keep the flux for all flavors in a single array with dimensions `(N_flavors,N_times,N_energies)` - this will make it easy to apply any transformation matrix, by just using matrix multiplication

So, something like this (pseudocode):

```python
class NeutrinoFlux:
    def __init__(self, data: Quantity[1/s/MeV], axes: Dict[str, Quantity]):
                 """ axes could be a dict{
                     'Flavors': list[Flavor], 
                     'E': Quantity[MeV],
                     'T': Quantity[s] }
                 """
       self.data = data
       self.axes = axes
        
    def apply(self, flavor_xform: FlavorTransformation) -> NeutrinoFlux:
        #apply the given mixing an return the transformed flux
        flux_transformed = flavor_xform @ self.flux #we can use matrix multiplication
        return NeutrinoFlux(flux_transformed, self.axes) 

```

## Benefits
1. Apply mixing having only `FlavorTransformation` and `NeutrinoFlux` classes (not coupled to `SupernovaModel`) - which is logical, because mixing acts on the neutrino signal rather than the supernova itself. 
2. Pass the `NeutrinoFlux` to the SNOwGLoBES/SimpleRate calculation directly. 

So instead of (or as alternative to) calling (pseudocode)
```python
files = generate_time_series(model, transformation, distance, E, t, ...)
results = simulate(files, detector, ...)
```
one could do
```python
flux = model.get_initial_flux(distance, E, t)
flux = flux.apply(transformation) 
results = rate_calculator.run(flux, detector, ...)
```
and have access to all the intermediate results IN MEMORY without reading additional files

Sheshuk:
Closes #214 

Initial tasks (before finalizing the interface)
- [x] Implement a flux/fluence/etc container class
- [x] `SupernovaModel.get_flux` should produce `Flux` class object 
- [x] Implement rate calculation using this class (passing Fluence to `SimpleRate` or similar class)
- [x] Implement `save/load` functions for this class (to do the caching, if desired) 

Next tasks:
- [x] Finalize interface
- [x] Document the new classes
- [x] Update SNEWPY documentation
- [x] Add tests
--------

#### Implement `Flux.apply(flavor_xform:FlavorTransformation)` method, to apply the transformation, and produce a transformed flux

Postponed till #242 is implemented, right now `SupernovaModel.get_flux` gets a `FlavorTransformation` argument to have transformed flux)

#### Merge `RateCalculator` and `SimpleRate` classes

I separated the common part as `SnowglobesData` base class, which handles the reading and accessing stuff from the snowglobes. So I think there is no need for this merge


Sheshuk:
Applying `FlavorTransformation` to the flux should wait until we implement #242.
At the moment, one can get the transformed flux via method `SupernovaModel.get_flux` providing the `flavor_xform` parameter there.

Sheshuk:
I've added a [notebook](https://github.com/SNEWS2/snewpy/blob/275a2b2ca8221e0648fcfc24e392b8924f7f1cb3/doc/nb/FluxContainer_demo.ipynb) describing the usage and main features of the new flux.Container class.

Sheshuk:
The base functionality is implemented, and before I continue, I'd like to have opinions on the interface I propose.
Please check the [jupyter notebook](https://github.com/SNEWS2/snewpy/blob/275a2b2ca8221e0648fcfc24e392b8924f7f1cb3/doc/nb/FluxContainer_demo.ipynb) with the usage examples of the implementation, and the code as well.
Any suggestions are welcome, please let me know if I'm moving to the right direction or not.

* For the rate computation: I separated new functionality in the `RateCalculator` subclass of `SimpleRate`. It is done temporarily for clarity (to avoid having many `run` and `compute_rates` methods in one class), we'll probably want to merge these classes into one later.
* In the rate computation, I used the integration over energy, instead of using bin-center approximation, so it should be slightly more precise.
I still implemented the old method as `RateCalculator.run_simple`
* A question: SimpleRate provides us with four tables of event rates: `smeared`,`unsmeared`,`weighted`,`unweighted`. Do we ever need to use `unweighted` variant? 
It looks like an intermediate calculation step, before multiplication by SNOwGLoBES weights (for each channel), so maybe it doesn't have any sense for the final user?

python/snewpy/flux.py
```diff
@@ -0,0 +1,141 @@
...
+#some standard container classes that can be used for 
+Flux = ContainerClass(u.one/u.MeV/u.s/u.cm**2, "d2FdEdT")
+Fluence = ContainerClass(Flux.unit*u.s, "dFdT")
+Spectrum= ContainerClass(Flux.unit*u.MeV, "dFdE")
+IntegralFlux = ContainerClass(u.one/u.MeV/u.s/u.cm**2, "F")
+
+DifferrentialEventRate = ContainerClass(u.one/u.MeV/u.s, "d2NdEdT")
+EventRate = ContainerClass(u.one/u.s, "dNdT")
+EventSpectrum = ContainerClass(u.one/u.MeV, "dNdE")
+EventNumbers = ContainerClass(u.one, "N")
```
JostMigenda, line 141:
```suggestion
#some standard container classes that can be used for 
Flux = ContainerClass(u.one/u.MeV/u.s/u.cm**2, "d2FdEdT")
Fluence = ContainerClass(Flux.unit*u.s, "dFdE")
Spectrum= ContainerClass(Flux.unit*u.MeV, "dFdT")

DifferentialEventRate = ContainerClass(u.one/u.MeV/u.s, "d2NdEdT")
EventRate = ContainerClass(u.one/u.s, "dNdT")
EventSpectrum = ContainerClass(u.one/u.MeV, "dNdE")
EventNumber = ContainerClass(u.one, "N")
```
* Mixup in Fluence/Spectrum names
* `IntegralFlux` seemed to be redundant to the `Flux` defined a few lines above.
* Typo in “differential”
* Make `EventNumber` singular, for consistency with other container classes

JostMigenda (commented):
Thanks for the great work @Sheshuk! The notebook is a nice illustration of the functionality and I think the overall direction is good.

There are a few conceptual issues I have with the `Container.sum` and `Container.integrate` functions.
Regarding `Container.sum`:
* Summing along the flavor axis is okay; though the use cases for that are fairly limited, I think.
* Summing along the time or energy axis of a `Flux` is physically meaningless: It gives results that depend on the binning in the original model file—especially so, since several model files have non-uniform binning, e.g. very narrow sub-ms time bins during the accretion phase and then wider time bins (up to ~100 ms) during the neutrino cooling phase. Simply summing those different bins without accounting for bin widths gives completely arbitrary numbers.
* Summing over the time axis of a `Fluence` or the energy axis of a `Spectrum` is meaningful and may be useful iff the integration that created that `Fluence`/`Spectrum` used >2 entries in `limits`.

Regarding `Container.integrate`:
* Integrating over the flavor axis doesn’t really make physical sense; and unfortunately, numerically it’s not the same as summing over the same flavours.
* Integrating along an axis that was already integrated should probably return an error.

I suspect all of these could be fixed by careful checks at the start of the `sum` and `integrate` functions; but I haven’t fully thought through the implementation details.

Other than that, I’ve suggested a few minor changes to the code inline.
Oh, and especially nice work on the `ContainerClass` helper function—that’s a very elegant way to solve this issue! 🤩

JostMigenda:
> * For the rate computation: I separated new functionality in the `RateCalculator` subclass of `SimpleRate`. It is done temporarily for clarity (to avoid having many `run` and `compute_rates` methods in one class), we'll probably want to merge these classes into one later.

For now, that’s definitely a good idea. Since this is intended to be temporary, I haven’t looked at the `RateCalculator` class to closely in my review above.

Before we can merge `RateCalculator` and `SimpleRate`, we’ll have to see whether we can update the implementation of `snewpy.snowglobes.calculate` to use this new flux class; but let’s leave that aside for now.

> * In the rate computation, I used the integration over energy, instead of using bin-center approximation, so it should be slightly more precise.
>   I still implemented the old method as `RateCalculator.run_simple`

Since `integrate` simply uses trapezoidal integration, does that really make a difference? I think it only matters at the edges of your integration region and only if those don’t align with bin centres, right?

> * A question: SimpleRate provides us with four tables of event rates: `smeared`,`unsmeared`,`weighted`,`unweighted`. Do we ever need to use `unweighted` variant?
>   It looks like an intermediate calculation step, before multiplication by SNOwGLoBES weights (for each channel), so maybe it doesn't have any sense for the final user?

Agree; the unweighted numbers are an intermediate step and I don’t see any use case for exposing them to the user.

Sheshuk:
@JostMigenda thank you for your review and suggestions!

About the integration/summation - I totally agree, in some cases it doesn't have physical sense. I thought that it's up to a user to choose physically meaningful operations, but having checks would be nice. I need to think about how to implement it. Probably keeping a set of `integrable` axes in the class?

> Since `integrate` simply uses trapezoidal integration, does that really make a difference? I think it only matters at the edges of your integration region and only if those don’t align with bin centres, right?

You're correct - if the integration bins are the same as the initial energy sampling points. But the sampling points can be much more detailed - this is especially the case I want for preSN models - while the integration bins are defined by the SNOwGLoBES smearing matrix bins.
In this case sampling cross-sections in the same energy points as flux, and integrating these in the needed energy bins give a more precise result.
I don't know if I describe it clearly.

Sheshuk:
I have:
1. Implemented the checks which axes can be summed, and which - integrated.
2. Separated the reading of SNOwGLoBES data files from `SimpleRate` to `SnowglobesData` class. 
Both `RateCalculator` and `SimpleRate` classes inherit from the `SnowglobesData`, and implement just the rate calculation in their own ways. 
So I would prefer not to merge `RateCalculator` and `SimpleRate`, and just keep the calculator in `snewpy.rate_calculator.RateCalculator`, and `snewpy.snowglobes_interface.SnowglobesData` separately, because the latter does not need to be exposed to user.
3. Postponed implementing the `Container.apply(transformation)`.
> Before we can merge `RateCalculator` and `SimpleRate`, we’ll have to see whether we can update the implementation of `snewpy.snowglobes.calculate` to use this new flux class; but let’s leave that aside for now.

So now most other tasks were finished or postponed, I think I'm ready to work on this.

JostMigenda:
Sorry about the delay in reviewing this!

Thanks for the updates on the summation/integration checks! And your other points seem fine to me, too. This looks almost ready to merge, there’s just two small points remaining:


First of all, the integration tests are currently failing due to an off-by-one error in the array dimensions introduced [in the last few commits](https://github.com/SNEWS2/snewpy/pull/236/files/784af50cce4edf9d3c9a1da2c202b92407f07596..c25c69ea65d38fabbd6d03b1fac888672ff68126); full traceback copied below. @Sheshuk, can you figure out what the issue is or should I take a look?
<details>
<summary><strong>Full traceback</strong></summary>
```
ERROR: test_simplerate (python.snewpy.test.simplerate_integrationtest.TestSimpleRate.test_simplerate)
Integration test based on SNEWS2.0_rate_table_singleexample.py
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/snewpy/snewpy/python/snewpy/test/simplerate_integrationtest.py", line 32, in test_simplerate
    snowglobes.simulate(SNOwGLoBES_path, tarredfile, detector_input=detector, detector_effects=True)
  File "/opt/hostedtoolcache/Python/3.11.3/x64/lib/python3.11/site-packages/snewpy/snowglobes.py", line 351, in simulate
    res=sng.run(flux_files, det)
        ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.3/x64/lib/python3.11/site-packages/snewpy/snowglobes_interface.py", line 307, in run
    result = self.compute_rates(detector, material, fluxes, energies)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.3/x64/lib/python3.11/site-packages/snewpy/snowglobes_interface.py", line 252, in compute_rates
    rates = np.dot(smear,rates) * effic
            ^^^^^^^^^^^^^^^^^^^
  File "<__array_function__ internals>", line 200, in dot
ValueError: shapes (200,200) and (199,) not aligned: 200 (dim 1) != 199 (dim 0)
```
</details>

And second, as discussed [previously](https://github.com/SNEWS2/snewpy/issues/214#issuecomment-1543888409) (see point 6), we should first use and carefully test these container classes ourselves before we make them part of the public API. Some of that will be in #251; and additionally in the pre-SN implementation. Until then, I would either omit these container classes from the public documentation completely (or at least add a warning box at the top of `flux.rst` that this API is still unstable).

Sheshuk:
Thanks for looking into this! 
> the integration tests are currently failing due to an off-by-one error in the array dimensions

Right, it was introduced in ee69beb13e64362e30f49c770b0d62011bb1e57d - it looks like we were treating the SNOWGLOBES binning incorrectly all this time, but the tests are still for the old version. 
I will try to revert it in this branch, so the tests pass, and will probably have to reimplement it again in #251

JostMigenda:
@Sheshuk and I have talked about this PR and #251 earlier today and fixed some remaining bugs. Since both PRs are closely related, it makes sense to first merge #251 (where these SNOwGLoBES binning issues & related tests are both fixed) into this one, and then merge this PR here into the main branch.

For the record, the final commit on this PR, before the first merge, was https://github.com/SNEWS2/snewpy/pull/236/commits/a42ac5438827d86756473bf0b4d707eaaf2b7e8a

And aside from the functional reasons for this PR, which we’ve discussed in detail, I’d just like to note that there are some nice performance benefits as well: In my testing using the example in `SNOwGLoBES_usage.ipynb` (`Zha_2017_s17` progenitor, `AdiabaticMSW` transformation, 60 time bins between 0.742 and 0.762 s), the total time for `generate_fluence`, `simulate` and `collate` went from ~35 s to ~10 s. ⚡

JostMigenda:
Some more issues with the integration tests. The first one was because of a tiny change in the keys for the results dictionary; that’s fixed now.

The current issue requires a bit more thought, because this PR changes the binning and that leads to slightly different event counts:

**Unsmeared:**
| Channel | Before | After | Change |
|--------|--------|--------|--------|
| ibd | 12498.485 | 12506.335 | 0.6 ‰ |
| nue_O16 | 110.328 | 110.435 | 1.0 ‰ |
| nuebar_O16 | 137.608 | 137.741 | 1.0 ‰ |
| nc | 275.077 | 275.397 | 1.2 ‰ |
| e | 1006.425 | 1006.915 |  0.5 ‰ |
| Total in Super-K¹ | 4488.94 | 4491.78 | 0.6‰ |

The unsmeared numbers typically change by less than 1 per mille; I think that’s nothing to worry about.

**Smeared:**
| Channel | Before | After | Change |
|--------|--------|--------|--------|
| ibd | 11916.581 | 11969.904 | 4.5 ‰ |
| nue_O16 | 109.625 | 109.772 | 1.3 ‰ |
| nuebar_O16 | 135.467 | 135.737 | 2.0 ‰ |
| nc | 28.176 | 28.208 | 1.1 ‰ |
| e | 455.939 | 461.574 | 12.2 ‰ |
| Total in Super-K¹ | 4046.65 | 4065.66 | 4.7 ‰ |

After smearing², the changes are much bigger, mainly in the electron scattering and IBD channels. In water Cherenkov detectors, those are the two channels which have the most events near the detector threshold (~5 MeV or so, depending on the detector); so any changes to the binning that cause events to slip above/below the detector threshold would predominantly affect those two channels.

So I think these changes are within what we’d expect; but maybe we should wait a little to let people comment? I’ll also try to double-check the binning code tomorrow, just to feel confident I didn’t miss anything.

¹ This is the sum of the rows above, multiplied by 0.32 (because the SNOwGLoBES detector config has 100 kt volume, while SK’s inner detector has 32 kt).
² smearing _and detector efficiency_, to be precise

evanoconnor:
@JostMigenda Are the percentages off by a factor of 10?

JostMigenda:
@evanoconnor Numbers are all in per mille, not per cent.

PR #237, opened 2023-05-16
Transpose the preSN flux arrays in , to match the (Ntime,Nenergy) shape, same as in CCSN

Sheshuk:
A small fix to make flux/initial_spectra, produced by preSN models have the same shape as in case of CCSN models: provided `time` and `energy` arrays, the shape of the of the result spectra should be `[len(time), len(energy)]`.

We don't have the checks for that resulting shapes in our tests. We'll need to revise the test suite  (not in this PR)

### DIFF
--- a/python/snewpy/models/presn.py
+++ b/python/snewpy/models/presn.py
@@ -67,7 +67,7 @@ class Odrzywolek_2010(SupernovaModel):
         alpha = np.expand_dims(alpha, 0)
         b = np.expand_dims(b, 0)
         fluence = a * Enu ** alpha * np.exp(-b * Enu) / (u.MeV * u.s)
-        result = {f: fluence * self.factor[f] for f in flavors}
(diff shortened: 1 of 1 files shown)


JostMigenda (approved):
LGTM!

(Just out of curiosity, is there any reason to prefer `(Ntime, Nenergy)` over `(Nenergy, Ntime)`? I assume it doesn’t really matter and we can just pick one convention as long as we’re consistent, right?)

Sheshuk:
At the moment it doesn't make any difference - because in all the places like `generate_time_series` we use python loop over the times, generating time points one by one:
https://github.com/SNEWS2/snewpy/blob/9096246377102e9d0e850c11396fa1e62779d642/python/snewpy/snowglobes.py#L110C2-L122

But it will make difference when we'll start using the vectorized computation in the future, so I decided to make the shapes uniform the same way as most of the CCSN models

PR #239, opened 2023-05-16
Removed duplicate times from Zha_2021 models

JostMigenda:
See discussion in #211.
(@sgriswol, @evanoconnor: It looks like we forgot to follow-up on this last autumn. I just re-discovered this, so I’ll open the PR now; hope that’s okay.)


mcolomermolla (approved):
This changes were already approved for other models, can be directly merged.

Issue #238, opened 2023-05-16
Don’t download model files for every test run


JostMigenda:
Right now, tests on GitHub Actions need to download all model files for every test run. This has multiple disadvantages:
* It breaks tests on PRs that add new model files, see e.g. #231 and [in particular this comment](https://github.com/SNEWS2/snewpy/pull/231#issuecomment-1317391415)
* It means tests running on GitHub may behave subtly different than on a developer’s local computer (where the model files are likely already downloaded and available in `snewpy.model_path`)
* Needing to download 100s of MB of model files slows down tests and wastes energy and bandwidth.

At least for model files that are in the GitHub repository anyway, the files are already available on the test runner; it would be sufficient to change `snewpy.model_path` to point to the `models/` directory in the repo. We could then also eliminate a lot of duplicate testing code in `tests/test_0{1,2}*.py`.
(Instead, we would probably need one download test running as part of the release workflow—after tagging, but before uploading the release to PyPI.)

Open questions: How to deal with models whose files are not in the GitHub repo? (E.g. on Zenodo?)

Sheshuk:
Hi Jost, this looks like a good solution for the time being.
But I thought the idea in the long run was to eventually move the data files to Zenodo or elsewhere, no?
Currently having the data strictly tied to the current release of SNEWPY on github leads to problems like this - when we change the data, but have no possibility to work with it properly, since it is not in the release yet.

I think moving all the data to an outside repository/Zenodo entry would solve all of the mentioned problems. 
Also developer won't have to download the data that is not needed, and cloning SNEWPY will be fast. And when the data files are updated, we can just make a new release of the data repository, decoupled from the development of the software

JostMigenda:
Moving the data files elsewhere would solve the first point; but not the second and third one. (In particular, having the files outside of GitHub would potentially make the third point much worse—downloading files over the internet from e.g. Zenodo is much slower than downloading them from another GitHub server that’s potentially a few meters away in the same data centre.)

For adding new models, yes, being able to use them directly from Zenodo (where modeling groups have hopefully made them available themselves) would be good. For existing model files, I’m not sure whether it’s worth the effort to move them elsewhere. But as you say, that’s a discussion for the long run; if we agree that the proposal above is a good solution in the meantime & since it doesn’t impede other long-term solutions, I think that’s good enough for now.

JostMigenda:
[...]

PR #247, opened 2023-05-20
Adding default option of downloading files from given URL, instead of fixed github one

Sheshuk:
Following my [suggestion](https://github.com/SNEWS2/snewpy/issues/238#issuecomment-1551981785) on making a separate repository for the data files, I would like to add the capability of downloading the files from any given url.

URLs can be formatted with any parameters, so I implemented the same github downloading in a new format for illustration:


python/snewpy/_model_downloader.py
```diff
@@ -113,7 +113,7 @@ def from_zenodo(zenodo_id:str, model:str, filename:str, path:str=model_path):
 
     Returns
     -------
-    file : FileHandle object.
+    file_url, md5sum
     """
     zenodo_url = f'https://zenodo.org/api/records/{zenodo_id}'
     path = Path(path)/str(model)
```
JostMigenda, line 119: Cleanup: Can you remove the remaining references to `path` in the docstring and function body?

JostMigenda: (and same in the `from_github` docstring)

JostMigenda (changes requested):
Some minor cleanup; otherwise this looks fine. Once we have a separate models directory and have tested that this works, I think we can remove the `from_github` function, which is just a duplicate of the default case, right?

Sheshuk:
I removed the `from_github`, cleaned up the docstrings, and also added a default parameter `snewpy_version` which can be put to the URL template. So it will always point to the current version.

python/snewpy/_model_downloader.py
Sheshuk: done

JostMigenda (approved):
LGTM, thanks @Sheshuk!

PR #253, opened 2023-06-02
Making a table with expected event rates - for cross-checks

Sheshuk:
Previously we had hardcoded values in the [python/snewpy/test/test_05_snowglobes.py] to compare the expected event numbers.
I've made a simple script to produce these event numbers into a file. The integration test now should compare the calculated total event rates to the table values.


mcolomermolla (approved):
I like having the dictionary with the table for the cross-checks. No comments on the PR changes, it can go ahead.

PR #254, opened 2023-06-02
Release version 1.4b1

### LINKED ISSUE #240: `Walk_2019` has outdated version in `model_files.yml`
In `model_files.yml`, `Walk_2019` currently lists `release_version: 1.2`. However, those files were updated in #211 which was released as part of v1.3, so we should update the yml file.

More generally, I think this `release_version` property is unnecessary; we always want to use the version of the model files that matches the SNEWPY version. Instead of fixing this one entry manually, we should see whether we can avoid manually specifying the version at all.

_(Thanks to @sgriswol and another IceCube collaborator for discovering this issue!)_

JostMigenda:
This fixes #240 (at least for now; a more permanent solution could be part of a separate PR, as mentioned there) and makes a beta release available on PyPI that doesn’t require a manual SNOwGLoBES install.


Sheshuk (approved):
LGTM. 
But is there any reason that we have different versions (`1.3` and `1.4b1`) for different models?

JostMigenda:
No particular reason. I just changed it to 1.3 for all models initially; then updated to 1.4b1 for all models that had changed since then.

PR #258, opened 2023-06-16
Flavor Transformations plus Earth matter

jpkneller:
A rewrite of the flavor transformation classes. 
1) Includes a new method called get_probabilities which returns a matrix of transition probabilities. This method is meant t replace the 8 prob_xy methods in the previous versions of the classes 
2) To avoid duplicating code, there are now two new intermediate classes called ThreeFlavorTransformation and FourFlavorTransformation between the abstract base class FlavorTransformation and the various flavor transformation classes. 
3) The ThreeFlavorTransformation class has two methods: one called vacuum and the other Earth_matter. These methods return the elements of the D matrix previously calculated in the flavor transformation classes. The Earth_matter method does an earth-matter effect calculation using the Sqa3Earth module. Compiling that module is described elsewhere. The FourFlavorTransformation class only has the vacuum method.  
4) The flavor transformation classes now have an optional argument to the constructor called AltAz. This is an astropy object which gives the altitude and azimuth of the supernova relative to the observer / detector. If not None, the earth matter effect will be done (if the altitude is less than zero) 
5) The previous version of the flavor transformation classes had two optional arguments to the constructor called mixing_angles and mass hierarchy. These have been replaced with the optional argument called mix_params which must be an instance of either the MixingParamters3Flavor or MixingParamters4Flavor classes in neutrino.py. If None, the default is to the use the normal mass hierarchy as before.


Sheshuk:
Hi @jpkneller! 
The SQA looks like an entire python package (interfacing the C++ code), with all the needed setup scripts.
Maybe it would make sense to put it into a separate repository, and publish to PyPI, and introduce it to SNEWPY as an external dependency?
This would make it easier for the user (no need for separate installation instructions in the readme etc.)

Sheshuk:
Sorry, I tried to merge the main branch and got it mixed up, reverting

jpkneller:
I am open to that idea.

jpkneller:
The Earth-matter effect module is now a separate repo and is not required for SNEWPY to work.

JostMigenda:
Covered in #351.

PR #263, opened 2023-09-01
Preparations for Python 3.12

### LINKED ISSUE #255: Remove distutils usage and update build process
The distutils module is [going to be removed in Python 3.12](https://docs.python.org/3.12/whatsnew/3.12.html#removed), to be released this autumn. Since we’re still using it in setup.py, it’s now time to update that.

And since invoking setup.py directly is also “[effectively deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html)”, I’ll see if I can update our build process more thoroughly in the process.

JostMigenda:
This PR removes dependencies on `distutils`, which is going to be removed in 3.12, when it is released in October. (Resolves #255.)

I’ll also add 3.12 to the testing matrix in the next commit; but those tests will likely still fail right now, since some dependencies don’t support 3.12 yet.


JostMigenda:
Within a week of the 3.12 release all of our dependencies are compatible, which is pretty fantastic! 😄 Time to mark this PR as ready for review.

mcolomermolla (approved):
This PR is ready to be merged.

PR #268, opened 2023-09-21
Fixes #267: explicitly call the units decomposition on Container creation

### LINKED ISSUE #267: Units simplification bug in the RateCalculator
Bug in `astropy.units` https://github.com/astropy/astropy/issues/15200 sometimes leads to units not simplified after the rate calculation, resulting in units like `cm^2/cm^2` or `h/s`.

Sheshuk:
Call "decompose" each time a new Container is created - this should prevent situations like _"cm2/cm2"_ or _"h/min"_ units. 

**Note:** this reduces the units to SI bases. If we want, we can list standard SNEWPY units here, so everything will be converted to them instead.

### DIFF
--- a/python/snewpy/flux.py
+++ b/python/snewpy/flux.py
@@ -64,6 +64,9 @@ from enum import IntEnum
 from copy import copy
 from functools import wraps
 
+#list of units which will be used as units for decomposition inside the Container
+snewpy_unit_bases = [u.MeV, u.m, u.s, u.kg]
+
 class Axes(IntEnum):
     """Enum to keep the number number of the array dimension for each axis""" 
     flavor=0, #Flavor dimension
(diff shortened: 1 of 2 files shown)


JostMigenda:
Thanks for investigating this further. As I already said on the AstroPy issue: Very nice detective work!

Regarding this PR, one quick example:
```Python
>>> data = [1, 2, 3] << 1/u.MeV/u.s
>>> data.decompose()
<Quantity [6.24150907e+12, 1.24830181e+13, 1.87245272e+13] s / (kg m2)>
```
So, yes, I think we would need to provide a custom unit list for decomposition; otherwise this will be unintuitive.

Also, have you had a chance to look at the test failures?

Sheshuk:
Thanks! You're right, I didn't think about MeVs. Now using `[MeV, m, s, kg]` as the bases. Will this be enough for general use cases?
Also it fixes these obscure test failures.

PR #271, opened 2023-10-03
Add a SNEWPY logo

JostMigenda:
Following the discussion on Slack this summer, this PR adds a SNEWPY logo:
snewpy logo: The word 'snewpy' in a monospace font, with an explosion emoji inside the letter 'p'.

In addition to this baseline logo, which is included in the README and docs, there are two alternate logos, which can be used e.g. in talks when highlighting that you’re talking about late-time emission (after successful shock revival) or about failed, BH-forming supernovae:

alternate logo: Like the baseline logo, but with a larger explosion emoji.

alternate logo: Like the baseline logo, but with the hole of the letter 'p' filled in to be completely black.

Thanks to @santosmv for asking about a logo (which inspired me to come up with some ideas in the first place) and to @jpkneller, @habig and Barry for feedback and suggestions that led to these final versions!


README.md
```diff
@@ -1,5 +1,7 @@
 # SNEWPY: Supernova Neutrino Early Warning Models for Python
 
+![SNEWPY logo](/doc/source/snewpy-logo.png)
```
Sheshuk, line 3: The logo displayed in the README is rather large, it makes me scroll down to get to the text. 
Maybe we should scale it down:
```suggestion
<img alt="SNEWPY" src="/doc/source/snewpy-logo.png" height=100px/>
```

Sheshuk (changes requested):
It's nice to have logo, thanks Jost!
In addition to the logo size in the README, one more minor suggestion:

The white background of this logo looks good on the light theme, but weird on the dark. Maybe we can have a transparent background and a white outline around the logo (so the black text would not be lost on dark bg)?

README.md
JostMigenda: Good point, thanks! Added this now; I also set a max-width for very narrow screens/windows.

Sheshuk (approved):
LGTM! Thanks, Jost!

JostMigenda:
Good comments, thanks!
For the dark version, here’s a first mockup of how that could look:
dark-mode versions of all three logo variants: black font with single-pixel white outline on black background, with explosion emoji as before

I like it in principle, but there’s still a bit of work to do to get the details right; e.g. the inner circle on the "p" looks awkward on the baseline logo. To produce a single file with transparent background that looks right on both black and white backgrounds, I’d probably have to redraw the whole thing from scratch in a dedicated vector graphics editor; and I’m not sure when I’d have time for that.

There’s also the issue that dark mode doesn’t necessarily mean exactly black background; and black font with white outline on dark grey background looks off. So maybe getting it actually right would require two separate images (light mode: black font, no outline; dark mode: white outline only) and a [`<picture>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture) element with [`prefers-color-scheme: dark`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme)?
I care enough that I’d actually like to try that at some point, but that probably won’t be anytime soon; so I don’t think we should wait for that.

Sheshuk:
Thanks! These look quite nice 👍🏻 But I agree, it needs some more thinking and work, and we shouldn't postpone release because of this.

mcolomermolla:
I agree that we shouldn't wait for that, I am happy with merging the white background version as it is for now :)

Issue #269, opened 2023-09-25
Allow more levels in the ModelRegistry YAML configuration tree


Sheshuk:
## What we have now
All models are listed in `models` section of the `model_files.yml`: https://github.com/SNEWS2/snewpy/blob/e0e47f500185fe71c801a6495d9526cad74e830c/python/snewpy/models/model_files.yml#L8-L14

That's ok for the ccsn case, but for adding the preSN models I would like to have them in a separate list. 

## What I suggest
Make a tree like
```yml
models: 
    ccsn:  
        Bollig_2016: 
            repository: *snewpy 
        Fornax_2019:  
            repository: *snewpy 
    presn:
        ...
```
and make the `_model_downloader` search not only by the model name, but also its module (`ccsn` or `presn`).

This will keep more order than just having all the models in one list.

sybenzvi:
This sounds fine if we feel it's useful to track the different emission models.

If we're going this route, would it be useful to rename the new repository `snewpy-models-ccsn` to `snewpy-models` and create `ccsn` and `presn` top-level folders in that repo?

JostMigenda:
Note that we already have a (currently private) repo called `snewpy-models-presn`. I have no strong opinions on whether we want to merge those model files into the new repo or keep the two repos separate; but if we're worried about the file size in the longer term as we add more and more models, keeping them separate would help a bit with that.

PR #273, opened 2023-10-05
Update model downloader

Sheshuk:
Mostly refactoring: same functionality, but using a more object-oriented approach.
1) Mostly refactoring: keeps the same functionality, calling `snewpy._model_downloader.get_model_data`
2) A global variable `registry` is defined and reused, so no need to read `model_files.yml` every time a model requests a file - it is read only once.
3) New feature: implements #269. This is not actually used now, but will be useful later, when we include preSN

### DIFF
--- a/python/snewpy/_model_downloader.py
+++ b/python/snewpy/_model_downloader.py
@@ -21,10 +21,11 @@ from snewpy import model_path
 from snewpy import __version__ as snewpy_version
 
 import logging
-logger = logging.getLogger('FileHandle')
+
+logger = logging.getLogger("FileHandle")
 
 
-def _md5(fname:str) -> str:
+def _md5(fname: str) -> str:
     """calculate the md5sum hash of a file."""
     hash_md5 = hashlib.md5()
     with open(fname, "rb") as f:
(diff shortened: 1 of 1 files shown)


python/snewpy/_model_downloader.py
```diff
@@ -117,54 +116,84 @@ def from_zenodo(zenodo_id:str, model:str, filename:str):
...
+        -------
+        Path of downloaded file.
+        """
+        tokens = config_path.split('.')
+        config = self.models
+        for t in tokens:
+            config = config[t]
+        #store the model name
+        model = tokens[-1]
+        modconf = config
```
JostMigenda, line 166: Unless I’m missing something, this renaming is unnecessary—and potentially confusing, since the code below uses a mixture of both variable names.

python/snewpy/_model_downloader.py
```diff
@@ -117,54 +116,84 @@ def from_zenodo(zenodo_id:str, model:str, filename:str):
...
+        if repo == 'zenodo':
+            url, md5 = _from_zenodo(zenodo_id=config['zenodo_id'],filename=filename)
+        else:
+            #format the url directly
+            params = { 'model':model, 'filename':filename, 'snewpy_version':snewpy_version}
+            params.update(modconf) #default parameters can be overriden in the model config
+            url, md5 = repo.format(**params), None
+        localpath = Path(model_path)/str(model)
+        localpath.mkdir(exist_ok=True, parents=True)
+        fh = FileHandle(path = localpath/filename,remote = url, md5=md5)
```
JostMigenda, line 178:
```suggestion
        fh = FileHandle(path=localpath/filename, remote=url, md5=md5)
```
Slightly awkward whitespace here and in a few other places in this file. (Not necessarily introduced in this PR, I just happened to notice now.) If you’ve got a code formatter set up in your IDE, it’d be nice to run it over the code quickly while you’re fixing the above comment; but absolutely no worries if not!

JostMigenda (changes requested):
Just two small notes; this looks good otherwise. Thanks!

python/snewpy/_model_downloader.py
Sheshuk: You're right, it's not needed at all. Fixed.

python/snewpy/_model_downloader.py
Sheshuk: I did't use the formatter before, now I processed the code with `black`. If you think this is a problem (our code in general is rather nonhomogenious in formatting), we could make it a necessary step or a part of tests before merging PR.

JostMigenda: I don’t think it’s a significant enough problem to impose an extra hurdle on contributors. (I know you’re an experienced developer and assumed you’d have a formatter set up already, otherwise I wouldn’t have bothered you with that. Thanks for doing this!)
We could consider a workflow that automatically `black`ens code that’s merged to `main`?

JostMigenda (approved):

PR #274, opened 2023-10-05
Using new CCSN models repository

### LINKED ISSUE #250: Put model files in a separate repository
Right now all the model data files are bundled in this repository, despite the new model registry functionality, which allows us to put them elsewhere.

As I suggested in #238 we can move out all the model files to a separate repository `snewpy_models`, and manage them there.

Advantages:
-----
1. Cut new release of the `snewpy_models` when the data is changed, without having to wait for `snewpy` release.
2. `snewpy` repository will become very lightweight (no need to download all the models during the checkout, for example if you want to work on a specific feature, implement an interface to a new model etc.)
3. We can create Zenodo entries for the `snewpy_models`, and direct the downloader there, if needed.

I think it can be done quite fast and simple. 
I implemented #247 to avoid hardcoding the repository path, and use any provided URL, so it should be easy to point to the new `snewpy_models` repo.

Sheshuk:
Closes #250


sybenzvi:
This looks good. Should we also delete the model files in the model folders?

sybenzvi (approved):
The changes to the YAML file look good. Before merging the branch, should we also delete the model files in the `snewpy/models` subfolder?

Sheshuk:
> Before merging the branch, should we also delete the model files in the `snewpy/models` subfolder

Let's do this in a separate PR #294

Issue #280, opened 2023-10-09
Type Error in FlavorTransformation Notebook


EmmaTintinger:
Error:
TypeError: MixingParameters() got an unexpected keyword argument 'mh'

Location:
-FlavorTransformation cell [6]

Error trace:
TypeError                                 Traceback (most recent call last)
Input In [6], in <cell line: 5>()
      2 angles_nmo = mp_nmo.get_mixing_angles() + (10*u.deg,)
      3 xf_nmo = AdiabaticMSWes(angles_nmo)
----> 5 mp_imo = MixingParameters(mh=MassHierarchy.INVERTED)
      6 angles_imo = mp_imo.get_mixing_angles() + (10*u.deg,)
      7 xf_imo = AdiabaticMSWes(angles_imo)

TypeError: MixingParameters() got an unexpected keyword argument 'mh'

PR #282, opened 2023-10-09
Update Jupyter notebooks

### LINKED ISSUE #277: Value Error in SNEWPY Notebooks
Error:
ValueError: Invalid value 'None' provided for parameter direction, allowed value(s): [1, 2, 3]

Locations:
Tamborra_2014 cell [3]
Walk_2018: cell [3]
Walk_2019 cell [3]

Solution:
Add “direction=1” (or a different allowed value) as a parameter in the definitions that are returning the errors

Error trace in Walk_2018:
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [3], in <cell line: 1>()
----> 1 model = Walk_2018(progenitor_mass=15*u.solMass, rotation='slow')
      3 model

File ~/snewpy/python/snewpy/models/ccsn.py:58, in _warn_deprecated_filename_argument.<locals>.decorator(cls, *args, **kwargs)
     54     msg = ''.join(['Initializing this model with a filename is deprecated. ',
     55                    f'Instead, use keyword arguments {list(cls.param.keys())}. ',
     56                    f'See `{cls.__name__}.param`, `{cls.__name__}.get_param_combinations()` for more info.'])
     57     warn(FutureWarning(msg), stacklevel=2)
---> 58 return func(cls, *args, **kwargs)

File ~/snewpy/python/snewpy/models/ccsn.py:308, in Walk_2018.__new__(cls, filename, eos, progenitor_mass, rotation, direction)
    304 if filename is not None:
    305     # Metadata creation is implemented in snewpy.models.base._GarchingArchiveModel
    306     return loaders.Walk_2018(os.path.abspath(filename))
--> 308 cls.check_valid_params(cls, progenitor_mass=progenitor_mass, rotation=rotation, direction=direction)
    309 filename = f's{progenitor_mass.value:3.1f}c_3D_{rotation}rot_dir{direction}'
    311 metadata = {
    312     'Progenitor mass': progenitor_mass,
    313     'EOS': 'LS220'
    314 }

File ~/snewpy/python/snewpy/models/base.py:462, in _RegistryModel.check_valid_params(cls, **user_params)
    460         continue
    461     else:
--> 462         raise ValueError(f"Invalid value '{user_param}' provided for parameter {key}, "
    463                          f"allowed value(s): {allowed_params}")
    465 # Check Combinations (Logic lives inside model subclasses under model.isvalid_param_combo)
    466 if user_params not in cls.get_param_combinations():

ValueError: Invalid value 'None' provided for parameter direction, allowed value(s): [1, 2, 3]

JostMigenda:
This fixes #277, #278 and #280.

The first two commits are just updating example notebooks to account for additional models we’ve added in v1.4.
The third one changes the name of an optional parameter to `MixingParameters` from `mass_order` back to `mh` (which it was in v1.3 and earlier). I’d like to release a v1.4.1 ASAP, to avoid anyone else running into issues with this.

### DIFF
--- a/models/Tamborra_2014/Tamborra_2014.ipynb
+++ b/models/Tamborra_2014/Tamborra_2014.ipynb
@@ -39,7 +39,7 @@
    "source": [
     "## Initialize Models\n",
     "\n",
-    "To start, let’s see what progenitors are available for the `Tamborra_2014` model. We can use the `param` property to view all physics parameters and their possible values:"
+    "To start, let’s see what progenitors are available for the `Tamborra_2014` model. We can use the `param` property to view all physics parameters and their possible values. However, for this model, not all combinations of these parameters are valid, so we use the `get_param_combinations` function to get a list of all valid combinations:"
    ]
   },
   {
@@ -48,14 +48,15 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "Tamborra_2014.param"
+    "# Tamborra_2014.param\n",
+    "Tamborra_2014.get_param_combinations()"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We’ll initialise both of these progenitors. If this is the first time you’re using a progenitor, snewpy will automatically download the required data files for you."
+    "We’ll initialise two of these progenitors. If this is the first time you’re using a progenitor, snewpy will automatically download the required data files for you."
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


JostMigenda:
@EmmaTintinger Thanks for reporting these three issues! I’ve made a branch with fixes; please feel free to check out this branch and verify that they are resolved.

@Sheshuk I’m fairly certain renaming the parameter back to `mh` shouldn’t have any side effects; but can you review this please?

Sheshuk (approved):
Thanks for the quick fixes, Jost!
Personally I don't like `mh` for a variable name, as it doesn't give any info to the user. But I agree we should roll back for backward compatibility. 
Still I suggest that we review our public interface at some point in the future to make it more informative.

PR #285, opened 2023-10-14
Mori axion 2023

sybenzvi:
Add models with axionlike particle production, using the 2D simulations from [Mori+ 2023](https://arxiv.org/abs/2304.11360). Estimates of the PNS mass are included in the paper. The commit adds the `Mori_2023` models to the `snewpy` registry, includes unit tests, and functions with the [mori-axion-models-2023 branch](https://github.com/SNEWS2/snewpy-models-ccsn/pull/4) of `snewpy-models-ccsn`.


python/snewpy/models/ccsn.py
```diff
@@ -742,6 +742,51 @@ def __new__(cls, filename=None, *, progenitor_mass=None):
     # Populate Docstring with abbreviated param values
     __new__.__doc__ = __new__.__doc__.format(**_param_abbrv)
 
+
+class Mori_2023(_RegistryModel):
+    """Model based on 2D simulations with axionlike particles, K. Mori, T.  Takiwaki, K. Kotake and S. Horiuchi, Phys. Rev. D 108:063027, 2023. All models are based on the non-rotating 20 M_sun solar metallicity progenitor model from S.E. Woolsey and A. Heger, Phys. Rep. 442:269, 2007. Data from private communication.
+        """
+    param = {'axion_mass' : [0, 100, 200],
+             'axion_coupling' : [0, 2, 4, 6, 8, 10, 12, 14, 16, 20] }
```
Sheshuk, line 750: I think it would be better for the user to operate the dimensional input parameters, i.e. `u.Quantity` (the same we do with `progenitor_mass` in other models, for example), and not work numbers in our internal units.

Maybe we should make the input parameters dimensional here
```suggestion
    param = {'axion_mass' : [0, 100, 200]<<u.MeV,
             'axion_coupling' : [0, 2, 4, 6, 8, 10, 12, 14, 16, 20]<<(1e-10/u.GeV) }
```

python/snewpy/models/ccsn.py
```diff
@@ -742,6 +742,51 @@ def __new__(cls, filename=None, *, progenitor_mass=None):
...
+        cls.check_valid_params(cls, axion_mass=axion_mass, axion_coupling=axion_coupling)
+
+        if axion_mass == 0:
+            filename = 't-prof_std.dat'
+        else:
+            filename = f't-prof_{axion_mass}_{axion_coupling}.dat'
+
+        metadata = {
+            'Axion mass': axion_mass * u.MeV,
+            'Axion coupling': axion_coupling * 1e-10/u.GeV,
```
Sheshuk, line 781: And remove the multiplying by units here
```suggestion
            'Axion mass': axion_mass,
            'Axion coupling': axion_coupling,
```

python/snewpy/models/ccsn.py
```diff
@@ -742,6 +742,51 @@ def __new__(cls, filename=None, *, progenitor_mass=None):
...
+        axion_coupling: int
+            Axion-photon coupling, in units of 1e-10/GeV. Valid values are {axion_coupling}.
+        """
+        # Load from Parameters
+        cls.check_valid_params(cls, axion_mass=axion_mass, axion_coupling=axion_coupling)
+
+        if axion_mass == 0:
+            filename = 't-prof_std.dat'
+        else:
+            filename = f't-prof_{axion_mass}_{axion_coupling}.dat'
```
Sheshuk, line 777: And convert values in needed units here
```suggestion
            filename = f't-prof_{axion_mass.to_value("MeV")}_{axion_coupling.to_value("1e-10/GeV")}.dat'
```

python/snewpy/models/loaders.py
```diff
@@ -702,3 +702,83 @@ def get_initial_spectra(self, t, E, flavors=Flavor, interpolation='linear'):
...
+        self.metadata = metadata
+
+        filebase = os.path.splitext(os.path.basename(filename))[0]
+        if 'std' in filebase:
+            am, ac = 0, 0
+        else:
+            am, ac = [int(_) for _ in filebase.split('_')[1:]]
+
+        self.pns_mass = mpns[(am,ac)] * u.Msun
+        self.metadata['PNS mass'] = self.pns_mass
```
Sheshuk, line 745: Here we're looking up the PNS mass from the given table, parsing the filename to get `am`,`ac` - to fill the metadata. Would it make more sense to do this lookup not in the loader, but in the `ccsn.Mori_2023` using the input parameters (or values from metadata)?

Sheshuk (changes requested):
Nice work implementing these models!

As I commented in the text, I have two small suggestions: 
1. make the input parameters Quantities
2. move filling the metadata (in particular PNS mass) to the `ccsn.Mori_2023` from `loaders.Mori_2023`

python/snewpy/models/loaders.py
sybenzvi: @Sheshuk, this is a good point. I shifted the PNS table into the `Mori_2023` class in the `ccsn` module for better encapsulation and dealt with the complications resulting from adding units to all quantities.

sybenzvi:
Thanks for the review @Sheshuk, I implemented all your requested changes and I think this model is good to go.

Sheshuk (approved):
Thanks for implementing this! Now everything looks good to me

PR #286, opened 2023-10-22
Detector and DetectionChannel classes for rate calculator

Sheshuk:
This is a refactoring of the `RateCalculator` internals in a more OOP style.
I define new classes `Detector` and `DetectionChannel` which contain the configuration info, read from SNOwGLoBES, and capable of calculating the event rates.

This is an internal change, which does not change the existing interface.

However it defines a lower level interface, which enables the following use case:

## Use case
As a user I want to be able to run the event rate estimation on a custom detector, not the ones which are present in SNOwGLoBES. I want to be able to make a copy of the existing detector, but set different mass, efficiencies, etc, and run the rate calculation for this case.
Or define a new detector with my own parameters (providing the detection channels etc)

## Usage example

See [Detector_demo.ipynb](https://github.com/SNEWS2/snewpy/blob/Sheshuk/Detector_class_for_RateCalculator/doc/nb/Detector_demo.ipynb) for more expanded example.

```python
import astropy.units as u
from snewpy.models import ccsn

#calculate the neutrino flux
m = ccsn.Bollig_2016(progenitor_mass=11.2<<u.M_sun)
t = np.linspace(0,10,101)<<u.s
E = np.linspace(0,100,51)<<u.MeV
flux = m.get_flux(t, E, distance=10<<u.kpc)
#get the detector configuration
from snewpy.rate_calculator import RateCalculator
rc = RateCalculator()
#read the standard detector configuration
det  = rc.read_detector('icecube')

#make a modified copy of the detector configuration
from copy import copy
det1 = copy(det)
det1.mass = 10000<<u.kt #different mass
det1.channels['ibd'].efficiency=1. #100% efficiency for IBD

#function to plot the rates
def plot_rate(rate, *args, **kwargs):
    x = rate.energy.to_value('MeV')
    y = rate.integrate_or_sum('time').array.squeeze()
    if rate.can_integrate('energy'):
        plt.plot(x,y, *args, **kwargs)
    else:
        plt.stairs(y,x, *args, **kwargs)

#calculate and plot right away
plot_rate(det.run(flux)['ibd'], label='scint20kt default', ls='-',color='k')
plot_rate(det1.run(flux)['ibd'], label='scint10kt modified', ls='--',color='k')
plt.legend()
plt.show()
```

## Outlook

In the future this can be a more convenient interface for the rate calculation: access and separate classes for Smearing and Efficiencies will allow to define them from the code more in a more flexible way.

## TODO:

- [x] Document the classes
- [x] Finalize interface (class/function names etc)
- ~Add tests!~ This functionality is mostly covered by the SNOWGLOBES tests, I don't see any need for unit tests of the inner interface now.
- [x] Add example notebook

Allow defining smearing matrices for: 
- [x] Gaussian smearing (reconstruction resolution)
- [x] Energy threshold
- [x] Monoenergetic response


python/snewpy/rate_calculator.py
```diff
@@ -13,11 +13,281 @@
...
+        """
+        rate = self._calc_interaction_rate(flux)
+        if apply_smearing:
+            if self.smearing is not None:
+                rate = self.smearing.apply(rate)
+        if apply_efficiency:
+            if isinstance(self.efficiency, FunctionOfEnergy):
+                rate = self.efficiency * rate
+            else:
+                rate = rate*self.efficiency
```
mcolomermolla, line 229: Here both `if` and `else` cases give same rate, is this condition needed?

python/snewpy/rate_calculator.py
```diff
@@ -13,11 +13,281 @@
...
+    """Description of a single detection channel in the detector"""
+    
+    def __init__(self, flavor:Flavor, xsec:callable, smearing:SmearingMatrix=None,
+                 efficiency:FunctionOfEnergy=1., weight:float=1.):
+        """
+        Parameters
+        ----------
+        name:str
+            channel name
+        flavor:Flavor or list[Flavor]
```
mcolomermolla, line 158: I am not 100% convinced that one DetectionChannel is defined by one flavor, that's the case of IBD but not of scattering/NC channels, where all flavors are involved. I guess in that case if I want the total for (e.g. pES), I should just define four detection channels (nue, anti-nue, nu_x, anti-nux) and sum them after right?

python/snewpy/rate_calculator.py
```diff
@@ -13,11 +13,281 @@
...
+        rateS = Container(rateS_array,
+                          flavor=rate.flavor, 
+                          time=rate.time, 
+                          energy=self.bins_smeared,
+                          integrable_axes=rate._integrable_axes)
+        return rateS
+
+    @classmethod
+    def from_Gaussian(cls, bins_true:u.Quantity, bins_smeared:u.Quantity,
+                      mean:Callable, sigma:Callable=1<<u.keV):
```
mcolomermolla, line 75: - Why not give also initial value for mean?
- Why the unit for sigma is in keV here when later you use MeV (line 115)?

python/snewpy/flux.py
```diff
@@ -265,10 +265,10 @@ def integrate(self, axis:Union[Axes,str], limits:np.ndarray=None)->'Container':
         if limits is None:
             limits = u.Quantity([xmin, xmax])
         limits = limits.to(ax.unit)
-        limits = limits.clip(xmin, xmax)
+        #limits = limits.clip(xmin,xmax)
         #compute the integral
         yc = cumulative_trapezoid(self.array, x=ax, axis=axis, initial=0)
-        _integral = interp1d(x=ax, y=yc, fill_value=0, axis=axis, bounds_error=False)
+        _integral = interp1d(x=ax, y=yc, fill_value='extrapolate', axis=axis, bounds_error=False)
```
mcolomermolla, line 271: I think it is better to make the calculation only in the time window where the initial model flux file is computed and avoid extrapolation, but this should have negligible impact

python/snewpy/rate_calculator.py
Sheshuk: Mean value is the average energy response `E_vis` for given neutrino energy `E_nu`. And it strongly depends on the process: for IBD `E_vis=E_nu - 784*u.keV`, while for NC on C12 it's just `E_vis = 15.1*u.MeV`

For sigma: 1 keV here is here as a very small default value, so if user doesn't provide the sigma, they get practically no smearing. And the units conversion is done by astropy.units so it works as expected.

python/snewpy/rate_calculator.py
Sheshuk: Oops, the dosctring is wrong, thanks for pointing it out!
I agree, it'd be better to have a single channel with several flavors. But this DetectionChannel class was made to represent 1to1 the channels in the SNOwGLoBES, and they are separate for each neutrino flavor:
```
nc_nue_C12
nc_nuebar_C12
nc_numu_C12
nc_numubar_C12
```

We can implement this (many flavors in one channel) as an additional feature, if needed, in a separate PR.

python/snewpy/rate_calculator.py
Sheshuk: You're right! That was initially done because `rate` contains a 3D array (Flavor,Time,Energy) and multiplication order matters, if efficiency is a vector (`rate*efficiency` will work correctly, on the Energy axis), and for the `FunctionOfEnergy` it's the class method `__mul__` which does all the work. 
But it'll be better if I get rid of this condition here, and for all cases it will be `rate = rate*self.efficiency`.

Below is just explanations of what I did in the commits, to achieve this:

1. Created `FunctionOfEnergy.__rmul__` method, so that it can be multiplied from the right.
2. Restricted the `flux.Container.__mul__` mehtod: if it's multiplied not by scalar of NumPy array, this returns `NotImplemented`. That means if I multiply by `FunctionOfEnergy` if will not use `Container.__mul__`, but instead `FunctionOfEnergy.__rmul__`

Sheshuk:
I think I addressed all the issues, could you review again @mcolomermolla?

python/snewpy/rate_calculator.py
mcolomermolla: okey, it is more clear now, thanks for explaining

mcolomermolla (approved):
I agree with all your last implementations, let's merge it.

PR #291, opened 2023-11-24
Update the way model downloader is used in the loaders

### LINKED ISSUE #290: Uninformative error messages, when loading models from wrong paths
When running models' loaders with the wrong path, we get all kinds of exceptions, but not the `FileNotFoundError`. 
For example:
```python
from snewpy.models import loaders

loaders.Tamborra_2014('/path/to/models/s15')
```
raises `InconsistentTableError`

and 
```python
loaders.Fornax_2021('/path/to/models/s15')
```
raises `IndexError: list index out of range` (because it failed to parse the filename, splitting it to tokens).

## Suggestion

We should make sure that the models raise `FileNotFoundError` when the path is wrong.

Sheshuk:
After #273, #287 (consider this only after those PRs)
Closes #290

1. Moved `GarchingArchiveModel` from `snewpy.models.base` to `snewpy.models.loaders`, because it handles loading model from file. `base` should be kept for general things.
2. Created classes `LocalFileLoader` and `RemoteFileLoader` in the `_model_downloader` - both implement the `_get_file(filename:str)->Path` method.


Sheshuk:
Closing this, as it was implemented in #296

PR #294, opened 2023-12-04
Remove model files from SNEWPY

Sheshuk:
As we moved all model files to separate [repository](https://github.com/SNEWS2/snewpy-models-ccsn), remove the data files from SNEWPY.


sybenzvi (approved):
This change looks good to me. I'd like to see some of the notebooks going into the `doc/nb` folder where they will be easy to find but we can discuss in a separate issue.

PR #296, opened 2023-12-12
Implementing preSN registry models

### LINKED ISSUE #243: Integrate preSN models with the models registry
Currently preSN models can only be instantiated by passing a path to model data.
Need to implement initialization via parameters, using models registry

### LINKED ISSUE #290: Uninformative error messages, when loading models from wrong paths

Sheshuk:
Closes #243 
Closes #290 
* Fix model tests which were not using absolute file paths (`test_02_models.py`)
* Moved GarchingArchiveModel from `base` to the `loaders` module.
* Created new `LocalFileLoader` class for checking if the requested file exists
* And `RegistryFileLoader` to download the requested file from the `_model_downloader`

### DIFF
--- a/python/snewpy/_model_downloader.py
+++ b/python/snewpy/_model_downloader.py
@@ -198,6 +198,24 @@ class ModelRegistry:
 registry = ModelRegistry()
 
 
+class LocalFileLoader:
+    @classmethod
+    def request_file(cls, filename:str)->Path:
+        "Require that the provided filename exists locally"
+        path = Path(filename).absolute()
+        if not path.exists():
+            raise FileNotFoundError(path)
+        return path
+
+class RegistryFileLoader(LocalFileLoader):
+    _registry = registry
+    
+    @classmethod
+    def request_file(cls, filename:str)->Path:
+        "Request file from the model registry"
+        return cls._registry.get_file(cls._config_path, filename)
+        
+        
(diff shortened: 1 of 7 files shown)


JostMigenda:
I’ve finished looking through the code-level changes; they all look good to me. Thanks a lot, @Sheshuk!

Before I accept and merge this PR, I’d like to try some usage examples. I noticed that the demo notebooks (e.g. https://github.com/SNEWS2/snewpy-models-presn/blob/master/models/Patton_2017/Patton.ipynb) still use the old (filename-based) initalisation, which doesn’t work with this PR. Have you started updating those already, or should I go ahead, update them as I try them out and make a PR on the snewpy-models-presn repo?

JostMigenda (approved):
Nice work, thank you!

Issue #298, opened 2024-01-11
Issues with number luminosity computation at small energies


soso128:
To compute the neutrino number luminosity, the energy luminosity is divided by the neutrino energy itself (Fornax models) or the mean energy (pinched spectra). To avoid computational issues, the zero value of the neutrino energy is set to a smal value, epsilon. This treatment is associated with issues for the Fornax models and for pinched spectra.

1. Fornax models:

The definition of the epsilon for the neutrino energy E and for the binned log energies logEbin are inconsistent when E is not in MeVs. The impact of this inconsistency on the number luminosities is shown in the figures below for the Fornax_2021 model with a 12Msun progenitor (top = current version, bottom = bug fixed).

![number_luminosity_lum_spec_12M_r10000](https://github.com/SNEWS2/snewpy/assets/23661417/41acd333-4388-457d-8520-f574376e5853)
![number_luminosity_lum_spec_12M_r10000_bugfix](https://github.com/SNEWS2/snewpy/assets/23661417/9bac078c-68aa-40ff-a7ad-1de6e9bfb79c)

2. Pinched spectra

When both the neutrino energy and the mean energy are small, the number luminosity can take strange, epsilon-dependent values, as can be seen in the figure below for the Nakazato_2013 model with the Shen equation of state, a 20Msun progenitor, z = 0.02, and a revival time of 300ms.

![number_luminosity_nakazato-shen-z0 02-t_rev300ms-s20 0](https://github.com/SNEWS2/snewpy/assets/23661417/50623dc3-7bb3-412c-9a7b-e0bf3e4b20bf)

This issue can be fixed by setting the number luminosity at zero neutrino energy to always be 0 (as it should be for a pinched spectrum). You can see the effect on the neutrino luminosity below. 

![number_luminosity_nakazato-shen-z0 02-t_rev300ms-s20 0_bugfix](https://github.com/SNEWS2/snewpy/assets/23661417/7ef35bf0-6f2b-4e78-a896-3f484d663d86)

Sheshuk:
@soso128 Thanks for investigating and fixing this issue!

PR #299, opened 2024-01-12
Fix bug in treatment of the E=0 point for fluences

soso128:
Possible solution for [issue #298](https://github.com/SNEWS2/snewpy/issues/298)


mcolomermolla (approved):
I agree with this fix to the issue

PR #301, opened 2024-02-08
Allow scalar input for 't', 'E' and 'data' arguments of Container class

Sheshuk:
Fixing problem if a user passes a scalar value to container. 
Example:
```python
import numpy as np
import astropy.units as u
from snewpy.flux import Flux

# this was the only valid way before this fix: wrap everything in []
Flux(data = [1] << u.Unit('1/(MeV s cm**2)'), 
     flavor=[Flavor.NU_E],
     time=[1]<<u.s,
     energy=[1]<<u.MeV)

#now it's possible to pass scalars if there is only a single value or point
Flux(data = 1 << u.Unit('1/(MeV s cm**2)'), 
     flavor=Flavor.NU_E, 
     time=1<<u.s,
     energy=1<<u.MeV)

#they're converted internally to arrays. This produces 
# d2FdEdT (1, 1, 1) [1 / (MeV s m2)]: <1 flavor(0;0) x 1 time(1.0 s;1.0 s) x 1 energy(1.0 MeV;1.0 MeV)>
```

### DIFF
--- a/python/snewpy/flux.py
+++ b/python/snewpy/flux.py
@@ -125,6 +125,21 @@ class _ContainerBase:
         self.energy = u.Quantity(energy, ndmin=1)
         self.flavor = np.sort(np.array(flavor, ndmin=1))
         
+        #list all valid shapes of the input array
+        Nf,Nt,Ne = len(self.flavor), len(self.time), len(self.energy)
+        expected_shapes=[(Nf,Nt,Ne), (Nf,Nt-1,Ne), (Nf,Nt,Ne-1), (Nf,Nt-1,Ne-1)]
+        
+        #treat special case if data is 1d array
+        if self.array.squeeze().ndim==1:
+            #try to reshape the array to expected shape
+            for expected_shape in expected_shapes:
+                if np.prod(expected_shape)==self.array.size:
+                    self.array = self.array.reshape(expected_shape)
+                    break
+        #validate the data array shape
+        if self.array.shape not in expected_shapes:
+            raise ValueError(f"Data array of shape {data.shape} is inconsistent with any valid shapes {expected_shapes}")
+        
         if integrable_axes is not None:
             #store which axes can be integrated
             self._integrable_axes = set(integrable_axes)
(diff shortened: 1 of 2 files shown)


JostMigenda:
Even worse—before, you needed `data = [[[1]]] << u.Unit('1/(MeV s cm**2)')` (with triple brackets). I agree this is a nice improvement; thanks!

JostMigenda (approved):

JostMigenda (changes requested):
Actually, on second thought, one thing gives me pause:

In addition to giving scalars for all arguments, this would also enable using a mix of scalars and vectors. E.g.
```Python
#now it's possible to pass scalars if there is only a single value or point
Flux(data = [1, 1, 1] << u.Unit('1/(MeV s cm**2)'), 
     flavor=Flavor.NU_E, 
     time=1<<u.s,
     energy=[1, 2, 3]<<u.MeV)

Flux(data = [1, 1, 1] << u.Unit('1/(MeV s cm**2)'), 
     flavor=Flavor.NU_E, 
     time=[1, 2, 3]<<u.s,
     energy=1<<u.MeV)
```
Both would produce the same 3D data array, but the intended physics meaning is different. I wonder if this makes it too easy to introduce mistakes?

Sheshuk:
Thanks, you're right! I didn't think about that. That's definitely against what the user would expect.

Sheshuk:
I added:
1. Reshaping `data` array **if it's 1D** to one of the expected shapes.
2. Array shape validation vs. expected shapes. 

Maybe it still doesn't cover all the possible corner cases.
But in general this class is not supposed to be created by user, but rather produced in the `SupernovaModel.get_flux` etc...

But I need to think of more strict conditions, when to use this reshaping

JostMigenda (approved):
I’m not quite sure about corner cases either; but this certainly adds useful guard rails for the intuitive(-seeming) case. And for all other cases, one has to consciously think about the array shape anyway; so having no extra guard rails is probably fine.

Thanks!

PR #310, opened 2024-04-05
times broadcasting for Fornax_2021

### LINKED ISSUE #259: times broadcasting in `get spectra`
When getting the initial spectra of different models, if the user passes multiple time entries the `Fornax2021` model cannot handle it where the others (that I have tested) can.

minimal example to reproduce;

```python
from snewpy.models.ccsn import Bollig_2016, Fornax_2021

bollig = Bollig_2016(filename=snewpy_models_path+"Bollig_2016/s11.2c")
fornax = Fornax_2021(filename=snewpy_models_path+"Fornax_2021/lum_spec_13M_r10000_dat.h5")

print(fornax.get_initial_spectra(t=[10]*u.s, E=10*u.MeV, flavors=[Flavor.NU_E])[Flavor.NU_E])
print(bollig.get_initial_spectra(t=[10]*u.s, E=10*u.MeV, flavors=[Flavor.NU_E])[Flavor.NU_E])

print(bollig.get_initial_spectra(t=[10, 12]*u.s, E=10*u.MeV, flavors=[Flavor.NU_E])[Flavor.NU_E])
print(fornax.get_initial_spectra(t=[10, 15]*u.s, E=10*u.MeV, flavors=[Flavor.NU_E])[Flavor.NU_E])
```

Multiple `t`'s work for Bollig but not for the Fornax model. 

The `snewpy` version I use is `snewpy==1.3`

This issue was also mentioned in #221

JostMigenda:
Fixes #259.
Additionally fixes a bug I discovered during this work, where for `interpolation="nearest"` a scalar value for E (e.g. `10*u.MeV`) would crash and a length-1 array (e.g. `[10]*u.MeV`) was required instead.

In addition to the usual tests, I’ve verified manually that the outputs for a few example t/E values were unchanged before/after this change. Additionally, I ran some performance checks, comparing the old and new versions with the following code snippet:

```Python
from snewpy.models.ccsn import Fornax_2021
import numpy as np

fornax = Fornax_2021(progenitor_mass=Fornax_2021.param['progenitor_mass'][1])
n_times = 100  # set this to 1, 10, 100 for benchmarking
times = np.linspace(1, 2, n_times) * u.s

# old: manually loop over times
for t in times:
    fornax.get_initial_spectra(t=t, E=list(range(100))*u.MeV, flavors=[Flavor.NU_E], interpolation="nearest")[Flavor.NU_E]

# this PR
fornax.get_initial_spectra(t=times, E=list(range(100))*u.MeV, flavors=[Flavor.NU_E], interpolation="nearest")[Flavor.NU_E]
```

Overall, this PR adds about ~10% overhead for a single time value, but for multiple time values it grows much less than linearly, so we’d see a performance boost for two or more time values, with about an order of magnitude better performance for 100 time values.

**Benchmark results for `interpolation = "linear"`**
Here, the length of the energies array passed in had no significant impact on results.

| `n_times` | old | this PR |
|--------|-----|-----|
| 1 | 0.62 ms | 0.69 ms |
| 10 | 6.2 ms | 1.2 ms |
| 100 | 61 ms | 5.5 ms |

**Benchmark results for `interpolation = "nearest"`**
Here, the range in each cell shows the impact of using an energy array of length 1–100 for a given `n_times`.
There’s likely some room for improvement here (as you’ll see in the diff, for one line I gave up and used a for loop over times, rather than spending even more time trying to figure out how to use NumPy broadcasting), especially for large arrays of energies and times; but this PR still a major improvement over the status quo and I don’t think it’s worth investing more time into this. (`interpolation = "nearest"` is a non-default value with little physics motivation; I doubt it’s used much.)

| `n_times` | old | this PR |
|--------|-----|-----|
| 1 | 0.55 – 0.65 ms | 0.59 – 0.70 ms |
| 10 | 5.1 – 6.2 ms | 0.97 – 2.0 ms |
| 100 | 51 – 61 ms | 4.3 – 15 ms |

### DIFF
--- a/python/snewpy/models/loaders.py
+++ b/python/snewpy/models/loaders.py
@@ -687,16 +687,17 @@ class Fornax_2021(SupernovaModel):
 
             elif interpolation.lower() == 'nearest':
                 # Find edges of energy bins and identify which energy bin (each entry of) E falls into
-                _logEbinEdges = _logE - _dlogE[0] / 2
-                _logEbinEdges = np.concatenate((_logEbinEdges, [_logE[-1] + _dlogE[-1] / 2]))
+                _logEbinEdges = _logE - _dlogE[0,0] / 2
+                _logEbinEdges = np.append(_logEbinEdges, np.expand_dims(_logE[:,-1] + _dlogE[:,-1]/2, 1), axis=1)
                 _EbinEdges = 10**_logEbinEdges
-                idx = np.searchsorted(_EbinEdges, E) - 1
-                select = (idx > 0) & (idx < len(_E))
+                idx = np.array([np.searchsorted(edges, E) - 1 for edges in _EbinEdges])
+                select = np.array([(_idx > 0) & (_idx < len(__E)) for _idx, __E in zip(idx, _E)])
 
                 # Divide luminosity spectrum by energy at bin center to get number luminosity spectrum
-                _dNLdE = np.zeros(len(E))
-                _dNLdE[np.where(select)] = np.asarray(
-                    [self._dLdE[flavor]['g{}'.format(i)][j] / _E[i] for i in idx[select]])
+                _dNLdE = np.zeros([len(j), len(np.atleast_1d(E))])
+                for i in range(len(j)):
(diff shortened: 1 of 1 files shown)


Sheshuk (approved):
Great work, @JostMigenda! It's a good solution with minimal changes.

PR #311, opened 2024-04-12
Update flavor_transformation.py

jpkneller:
I think there's a mistake in the formulae for P11 and P12 in the QuantumDecoherence class. They don't match what is found in equation II.4 of the dos Santos et al paper.

### DIFF
--- a/python/snewpy/flavor_transformation.py
+++ b/python/snewpy/flavor_transformation.py
@@ -1739,7 +1739,7 @@ class QuantumDecoherence(FlavorTransformation):
 
         :meta private:
         """
-        return 1/3 + 1/2 * np.exp(-(self.Gamma3 * (E/self.E0)**self.n + self.Gamma8 * (E/self.E0)**self.n / 3) * self.d) + 1/6 * np.exp(-self.Gamma8 * (E/self.E0)**self.n * self.d)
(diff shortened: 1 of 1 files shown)


JostMigenda:
@santosmv If you have a moment, can you check this and confirm whether the issue is here or in the paper?

Once that’s confirmed, the test failures are quite straightforward to fix; I can take care of this while I review & merge.

JostMigenda:
@jpkneller After a closer look at the paper, I noticed the line just under eqn. (II.6):
> leading to `\gamma_3 = \Gamma_3 + \Gamma_8 / 3` and `\gamma_8 = \Gamma_8`

It looks like that’s where this additional term is coming from?
(The notation with upper/lower case gammas easily leads to confusion; so even if the current code is correct, it might be useful to add a comment for clarity?)

santosmv:
Apologies for the notation. You're correct that the uppercase \Gamma_3 and \Gamma_8 originate from equations (II.5) and (II.6), respectively, and come from the decomposition of lowercase gammas into coefficients of V_p operators. This notation has two primary purposes: firstly, to derive parameters that remain independent of each other (in terms of V_p coefficients), and secondly, to simplify the comparison with results of another paper that introduced this notation (available at https://arxiv.org/abs/1909.09504).

JostMigenda:
No worries; I think the notation makes perfect sense in the paper. It’s just that code isn’t typeset as nicely, so legibility is necessarily impaired a bit.

It looks like this is resolved, so I’ll close it now.

PR #312, opened 2024-04-15
Dveyrat/extended cooling tail model

dveyrat:
Adds extended supernova model class which models neutrino luminosity during the cooling tail.


sybenzvi:
Before completing the review, let's close the loop with collaborators from Hyper-K on this feature (email forwarded by @JostMigenda).

JostMigenda:
Had a first look while waiting for my transfer in ORD. Boarding begins now, so just a few notes for next week:
* I’ll suggest a few more minor code changes later.
* Some usage examples would be good. I think we already have one in the v2.0 paper draft; but it would be good to have that available in the repo (e.g. as an example notebook or in the docstring)
* Are there any useful tests we could add?

python/snewpy/models/extended.py
```diff
@@ -0,0 +1,110 @@
...
+        gamma : float
+            Third cooling tail model parameter.
+
+        Returns
+        -------
+        astropy.Quantity
+            Integrated luminosity calculated from Ekanger et al. (2023) model.
+        """
+        logE = (m_PNS.value * alpha) + (SRT.value * beta) + gamma
+        return (10**logE) * u.erg
```
JostMigenda, line 110: If I understand correctly, this fit is specific to the `Nakazato_2013` models (e.g. due to the 20s time range and the shock revival time) and does not make sense for other models? The function also does not appear to be used anywhere, so would it be better to delete it?

python/snewpy/models/extended.py
```diff
@@ -0,0 +1,110 @@
...
+            Power law factor (default: -1)
+        A : astropy.Quantity
+            Normalization factor (default: None)
+        tau_c : astropy.Quantity
+            Exponential decay characteristic timescale (default: 36 s)
+        alpha : float
+            Exponential decay factor (default: 2.66)
+        """
+        for t in ts:
+            self.time = np.append(self.time, t)
```
JostMigenda, line 81: Should we enforce here that the times in `ts` are larger than `base_model.time[-1]`?

Or alternatively, if we’d like to include a larger time range in `ts` (e.g. to compare the “true” luminosity with the “extended” luminosity at early times), I think it would be better to keep the “true” and “extended” data points separate. If we just append the extended points to `self.time` (here) and `self.luminosity` (below), it’s more difficult to plot the true/extended luminosity separately and we get plotting artifacts, like in this example from the demo notebook:

image

JostMigenda (changes requested):
A couple of comments and fixes. They’re mostly minor, but some may require discussion.

python/snewpy/models/extended.py
```diff
@@ -0,0 +1,110 @@
...
+        """Get neutrino luminosity from supernova cooling tail luminosity model.
+
+        Parameters
+        ----------
+        t : astropy.Quantity
+            Time to evaluate luminosity.
+        k : float
+            Power law factor (default: -1)
+        A : astropy.Quantity
+            Normalization factor (default: 1e51 erg/s)
```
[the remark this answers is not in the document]
sybenzvi: Agreed, there is no good reason for the different default arguments. They are now the same in both member functions.

python/snewpy/models/extended.py
JostMigenda: I’d … probably err on the side of not making the user keep track of such things; but I can also see how it might be lead to confusion if we try to be “too smart” and drop entries when the user might not expect that. 🤔

Maybe we should keep it as is for now, merge, and ask people to test a beta release? And then we can still change it based on feedback.

python/snewpy/models/extended.py
```diff
@@ -0,0 +1,107 @@
...
+
+        self.__dict__ = base_model.__dict__.copy()
+        for method_name in dir(base_model):
+            if callable(getattr(base_model, method_name)) and method_name[0] != '_':
+                if method_name == 'get_initial_spectra':
+                    self._get_initial_spectra = getattr(base_model, method_name)
+                else:
+                    setattr(self, method_name, getattr(base_model, method_name))
+        self.t_final = self.time[-1]
+        self.L_final = {flv: self.luminosity[flv][-1] for flv in Flavor}}
```
JostMigenda, line 26:
```suggestion
        self.L_final = {flv: self.luminosity[flv][-1] for flv in Flavor}
```
Fix an unfortunate typo

sybenzvi:
Thanks @JostMigenda , we can figure out the interface later. After your merger I did update the `extend` function to only add times after the end of the model time, but I didn't test this yet. If you're satisfied we can punt on that until after the release.

Issue #185, opened 2022-02-09
Include notebooks in documentation


JostMigenda:
[As discussed elsewhere](https://github.com/SNEWS2/snewpy/issues/177#issuecomment-1034025333), it’s possible to [include Jupyter notebooks in Sphinx documentation](https://docs.readthedocs.io/en/stable/guides/jupyter.html).
I haven’t tried this out yet, but it looks like a great way to make all of our excellent notebooks (both on specific models and with general usage examples) more visible.

Sheshuk:
This is a good idea.
Apart from the documentation, they will be a good addition to our integration testing - checking that they work without errors before merging a branch will force us to keep them up to date.

sybenzvi:
I want to raise the prominence of this issue. I've noticed some model notebooks are now broken. In addition, we posted the notebooks without processed plots and code, largely for the benefit of developers. Unfortunately, this makes them less useful as a documentation resource, and once the notebooks are broken, we cannot go look on GitHub for how the plot are supposed to appear.

JostMigenda:
I’ll take a look later this afternoon.

Meanwhile, can you let me know which notebooks are broken? Since #322 we’re testing the model notebooks in CI on every PR, so that shouldn’t happen—would be good to figure out if there’s a gap in our testing strategy that we should close.

PR #313, opened 2024-04-22
Restore Model Notebooks

### LINKED ISSUE #288: clean up model notebooks and documentation
We are factoring the snewpy repo into code that lives here and model files that live in [snewpy-models-ccsn](https://github.com/SNEWS2/snewpy-models-ccsn). Currently the jupyter notebooks which demonstrate the models live in the folder
```
snewpy/models/{model_name}/{model_name}.ipynb
```
I propose we remove the top-level models folder and copy all the jupyter notebooks into
```
snewpy/doc/nb
```

JostMigenda:
Resolves #288. Once this is merged, we should remove the notebooks from the model repos; I’ll post separate PRs there in a moment.

This is also a first steps towards integrating notebooks into the online docs, see #185.


mcolomermolla (approved):
I think this can be merged straight away. Maybe just:
- notice that `Mori_2022` notebook has been modified for a bug fixed after the last commit here, so take into account before merging
- I would maybe call the `doc/nb/ccsn` -> `doc/nb/models` instead? but just a matter of taste

JostMigenda:
Thanks @mcolomermolla!

I’ve merged the Mori notebook changes manually; also did some very small cleanup of the typesetting in the example code.

The directory should be `doc/nb/ccsn/` in this case; note that we also have `doc/nb/presn/`. (It’s rather unlikely we’ll get identically named models for both ccsn and presn, but let’s play it safe.)

Issue #295, opened 2023-12-07
Issue with coherent scattering


mcolomermolla:
Trying to simulate the DS20k detector I encountered the following problems:

- Material is not defined for ds20 detector (that's not a real problem because one can give it directly to rate calculator), but maybe for the user it is not obvious.
- The coherent scattering channels are dropped as they have a different binning compared to others:
`
# drop coherent scattering channels, which have different binning
                df = df[df["name"].str.startswith('coh_') == False]
`

mcolomermolla:
> Thanks for writing these up!
> 
> I’ll start with your last two points:
> 
> * It looks like the efficiency files (e.g. [`effic_coh_helm_nue_Ar_ds20.dat`](https://github.com/SNOwGLoBES/snowglobes/blob/master/effic/ds20/effic_coh_helm_nue_Ar_ds20.dat) give an efficiency of `0.0` independent of energy; I suspect that’s why your rates are zero even after the modifications.
> * Yes—`snowglobes_data` only contains the efficiency files present in the SNOwGLoBES repo. In practice, I don’t think these missing files matter much, though: Since the visible energy in neutrino-electron scattering is O(MeV)—i.e. several orders of magnitude higher than typical signals these DM detectors are built for—I’d expect that the efficiency is close to 1, which is also the default value used by snewpy if no efficiency file is provided.
> 
> Once those two points are fixed (which are outside the scope of snewpy, really; probably something we should discuss with DS20k folks in SNEWS first, and then discuss/track in an issue on the SNOwGLoBES repo), I agree the first two points are things we should improve in snewpy.

Yes, I realised after that efficiencies are zero... so basically, deleting those files so that it uses by default eff=1 should be fine for the moment to get a reasonable output for the coherent channels. However, it still won't run all the channels together because of the different binning, I will have a look at that.

OliverMacfadyen:
Hi I am a PhD student on DarkSide and I ave just started working on snewpy and snews with a group at Oxford university. I was just wondering if anyone would be available to get me up to speed on this issue, where it is currently up to and if there is anything I can do to help.

mcolomermolla:
Hi Oliver,
Sorry for the late reply. At the moment, there have been no updates. If you want, I can write you an email and we set a zoom meeting so that I can explain to you the issues and where to start as we originally planned.

Sheshuk:
Fixed in #319

PR #319, opened 2024-04-25
cevns-fixes

mcolomermolla:
This PR removes the dropping of coherent channels and adds missing materials
This PR is a solution to issue #295 together with the PR by Jost in snowglobes.


python/snewpy/snowglobes_interface.py
```diff
@@ -113,8 +117,9 @@ def _load_channels(self, chan_dir):
                 # Format 2: explicit binning, differs per channel
                 tokens = l.split()[6:]
                 df = pd.read_table(f, delim_whitespace=True, index_col=1, comment='%', names=['name','n','parity','flavor','weight'], usecols=range(1,6))
-                # drop coherent scattering channels, which have different binning
-                df = df[df["name"].str.startswith('coh_') == False]
+                # drop coherent scattering channels if channel is argon_400bins, mixing channels which have different binning
```
Sheshuk, line 120: The comment is slightly misleading
```suggestion
                # drop coherent scattering channels if material is argon_400bins, mixing channels which have different binning
```
But do I understand correctly, that we skip some channels because in our data structure we have only one binning for each material, while Snowglobes can have separate binning for each channel? 
Probably this is something we should fix on our side.

Sheshuk (approved):
This looks good to me, thanks, @mcolomermolla! 

I'm just a little confused with skipping `argon_400bins` material because of binning. But probably this needs to be a separate issue.

python/snewpy/snowglobes_interface.py
JostMigenda: Good question! But either way, it’s certainly beyond the scope of this PR, so I moved this to issue #334 and added my thoughts there.

mcolomermolla: Yes, indeed I opened an issue in snowglobes: 

https://github.com/SNOwGLoBES/snowglobes/issues/24

But as Jost said indeed opnining it in snewpy seems more reasonable, and that was goes beyond the scope of this PR.

PR #325, opened 2024-04-30
Fix bulk downloader

JostMigenda:
One more problem discovered while preparing the v1.5 release: The bulk downloader (`snewpy.get_models()`) broke when moving model files to separate repositories.

This PR fixes it and ensures it uses the same download code as model initialisers (iterating over all parameter_combinations, rather than using its own dictionary of hardcoded file names); to avoid future breakage.

As a side effect, the bulk downloader will now download to the same `snewpy.model_path`, rather then the current directory. This is good for consistency and may avoid duplicate downloads.
The parameter `download_dir` was never officially documented; I’m still marking it as deprecated though, to be safe.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -27,11 +27,8 @@ jobs:
         TWINE_USERNAME: __token__
         TWINE_PASSWORD: ${{ secrets.PYPI_JM }}
       run: |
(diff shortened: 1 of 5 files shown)


Sheshuk (approved):
Looks good to me.
Now everything concerning the models' downloading is consistent, very good!

PR #328, opened 2024-05-03
Fix docs build and add it to CI workflow

### LINKED ISSUE #327: docs page fails to build
According to the badge displayed in the README file (see screenshot), ReadTheDocs currently fails to build our docs.

image

When building the docs locally, I notice there is a warning
`snewpy/python/snewpy/models/presn.py:docstring of snewpy.models.presn.Patton_2017:10: WARNING: Duplicate explicit target name: "zenodo".`
and RTD appears to treat warnings as errors in its builds. (Which is arguably a good practice.)

The intermediate fix to eliminate this warning is to turn the links into [anonymous references](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#hyperlink-references); but to avoid similar things happening again, we should also do the following:

- [ ] treat warnings as errors in our docs Makefile
- [ ] build docs as part of our CI workflows

I’ve started working on this & will try to get a PR out on Friday. Would be good to get this in ASAP (before releasing v1.5), so we have docs available as soon as the release is published.

JostMigenda:
Fixes #327.

In addition to fixing the warning itself, this includes measures to more quickly discover similar issues in the future:
This PR now treats warnings as errors in our docs Makefile, consistent with ReadTheDocs’ configuration. It also builds docs as part of our CI workflows, so we notice future issues on every PR and can fix them before merging broken docs into `main`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,9 +36,13 @@ jobs:
           python -m pip install --upgrade pip
       - name: Install SNEWPY
         run: |
-          pip install ".[dev]"
+          pip install ".[dev,docs]"
       
-      # Run the unit tests
-      - name: Test with pytest
+      - name: Run unit tests with pytest
         run: |
           pytest -m 'not snowglobes'
+
+      - name: Build HTML docs
+        run: |
+          cd doc/
+          make html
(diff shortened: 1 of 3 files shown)


Sheshuk (approved):
Looks good and works for me. Thanks!

PR #329, opened 2024-05-03
v1.5 release

JostMigenda:
Changes for the v1.5 release.

Note: Because of a bug in SNOwGLoBES v1.3.1 and earlier ([where `channels_argon.dat` would incorrectly refer to `_he` channels](https://github.com/SNOwGLoBES/snowglobes/pull/28)), SNEWPY would not find the efficiency files and assume a fixed efficiency of 100%.
With the fix in SNOwGLoBES v1.3.2, SNEWPY now finds the correct efficiency files, so the total event count for the `ar40kt` detector is about 5% lower than in previous versions.


.github/workflows/publish.yml
```diff
@@ -22,19 +22,26 @@ jobs:
...
+
+    - name: Check and get Version Number
+      id: get_version
+      run: |
+        pip install .
+        PYTHON_VERSION=`python -c 'import snewpy; print(snewpy.__version__)'`
+        echo "PYTHON_VERSION=${PYTHON_VERSION}"
+        GIT_VERSION=${GITHUB_REF/refs\/tags\/v/}
+        echo "GIT_VERSION=${GIT_VERSION}"
+        if [ $PYTHON_VERSION != $GIT_VERSION ]; then exit 1; fi
```
Sheshuk, line 34: I really like this versions check, nice!

Sheshuk (approved):
Looks good, thanks for this work!

PR #330, opened 2024-05-07
Add presn.Yoshida_2016 model

### LINKED ISSUE #215: Add presn.Yoshida model
Implement Yoshida model from https://zenodo.org/record/3778014

Sheshuk:
Closes #215 

- [x] Add loader class
- [x] Add registry model 
- [x] Add example notebook
- [x] Cross-check the resulting  plots with the model paper

### DIFF
--- a/python/snewpy/models/presn_loaders.py
+++ b/python/snewpy/models/presn_loaders.py
@@ -148,10 +148,9 @@ class Yoshida_2016(SupernovaModel):
             data = []
             T = []
             while (line := f.readline()):
-                if(not line):break
-                N_step,T_step,logT_c,logRho_c = [conv(token) for conv,token in zip([int,float,float,float],line.split())]
+                if not line: break
+                T += [float(line.split()[1])]
                 data += [[np.loadtxt(f, max_rows=100).flatten() for i in range(4)]]
-                T+=[T_step]
         times = np.array(T)
         energies = np.concatenate([
                 np.linspace(0,10,1001)[1:],
--- a/python/snewpy/models/presn_loaders.py
+++ b/python/snewpy/models/presn_loaders.py
@@ -158,6 +158,10 @@ class Yoshida_2016(SupernovaModel):
                 np.linspace(10,20,501)[1:]
             ])
         dNdEdT = np.stack(data, axis=1)
+        #rearrange flavors from NU_E, NU_E_BAR,_NU_X, NU_X_BAR to current 
+        indices = np.argsort([Flavor.NU_E,Flavor.NU_E_BAR, Flavor.NU_X, Flavor.NU_X_BAR]) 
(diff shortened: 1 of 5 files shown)


Sheshuk:
Cross-check with Yoshida_2016 [paper](https://arxiv.org/abs/1606.04915) (Fig.5):
image

Using a script here I plot the image 
https://gist.github.com/Sheshuk/3a11f9b033fb8a3029b53fada9aa4f9a
image

JostMigenda:
From the figure, it looks like the NU_E and NU_X rates are almost identical (both solid lines overlap); while the NU_E_BAR and NU_X_BAR are both almost identical to each other (both dashed lines overlap), and a factor of ~4 lower than the solid lines. However, physically, this doesn’t seem right at all: The main emission mechanism is pair production; so NU_E should be very similar to NU_E_BAR, NU_X should be very similar to NU_X_BAR, and there’s no a priori reason why NU_E and NU_X should be similar.

Therefore, I suspect that NU_E_BAR and NU_X have been mixed up here. Looking at the README, the input files use the order NU_E, NU_E_BAR, NU_X, NU_X_BAR; while the `Flavor` enum in `snewpy.neutrino` uses the order NU_E, NU_X, NU_E_BAR, NU_X_BAR. Maybe that’s the origin of the problem?

python/snewpy/models/presn_loaders.py
```diff
@@ -139,3 +139,36 @@ def get_initial_spectra(self, t, E, flavors=Flavor):
...
+    """
+    def __init__(self, filename):
+        with open(self.request_file(filename)) as f:
+            data = []
+            T = []
+            while (line := f.readline()):
+                if(not line):break
+                N_step,T_step,logT_c,logRho_c = [conv(token) for conv,token in zip([int,float,float,float],line.split())]
+                data += [[np.loadtxt(f, max_rows=100).flatten() for i in range(4)]]
+                T+=[T_step]
```
JostMigenda, line 154:
```suggestion
                if not line: break
                T += [float(line.split()[1])]
                data += [[np.loadtxt(f, max_rows=100).flatten() for i in range(4)]]
```

JostMigenda (changes requested):
Looks good to me now; but let’s add this to the `test_presn_rates.py` to have at least some basic checks that the model initializes successfully and produces reasonable rates.

python/snewpy/models/presn_loaders.py
Sheshuk: This adds a "magic number" `[1]` here, so for me it's less clear to see, what is going on.
But it's subjective. So if it makes code look better to you, let's do it

JostMigenda (approved):
Looks good; thanks a lot! ☺️

PR #331, opened 2024-05-09
Update ccsn_loaders.py

jpkneller:
I think there's an error in lines 60-63. It should be _flav.name not flavor.name


JostMigenda:
Could you explain why?

The model input files don’t distinguish nu_x and nu_x_bar, so `_flav` is used when constructing the filename to read the nu_x_bar flux from the nu_x file. But after reading the data, outside code shouldn’t have to care about whether the input files distinguish the two. So I think it makes sense to use `flavor` everywhere else, like we do now?

jpkneller:
Aah, I was looking at an old version. Forget this PR.

JostMigenda:
No worries!

PR #341, opened 2024-06-04
Fix Windows-specific bugs

### LINKED ISSUE #340: Warren_2020 downloads fail on Windows
A project student running snewpy on Windows discovered an error trying to download `Warren_2020` files; where trying to run the code from the sample notebook resulted in:
```
HTTPError: 404 Client Error: Not Found for url: https://github.com/SNEWS2/snewpy-models-ccsn/raw/v0.2/models/Warren_2020/stir_a1.23%5Cstir_multimessenger_a1.23_m10.0.h5
```

Note the `%5C` in the URL. This is [percent encoding](https://en.wikipedia.org/wiki/Percent-encoding#Character_data) for the `\` character (instead of the `/` character that would make this a valid URL).
I think I’ve got a fix ready; but I’ll also see if we can add CI tests under Windows, to avoid things like this in the future.

JostMigenda:
This should eventually fix #340.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -47,12 +47,25 @@ jobs:
           python -m unittest python/snewpy/test/simplerate_integrationtest.py
           pytest -m 'snowglobes'
       - name: Run Jupyter notebooks
+        if: ${{ ! startsWith(matrix.os, 'windows') }}
         run: |
           pip install nbclient ipykernel
           jupyter kernelspec list
           cd doc/nb/
           jupyter execute ccsn/*.ipynb
           jupyter execute presn/*.ipynb
+          jupyter execute AnalyticFluence.ipynb FlavorTransformation.ipynb
+          # Download necessary progenitor file, before executing the notebook
+          python -c 'from snewpy.models.ccsn import Zha_2021; from astropy import units as u; _ = Zha_2021(progenitor_mass=17*u.solMass)'
+          jupyter execute SNOwGLoBES_usage.ipynb
+      - name: Run Jupyter notebooks on Windows
+        if: ${{ startsWith(matrix.os, 'windows') }}
+        run: |
+          pip install nbclient ipykernel
+          jupyter kernelspec list
+          cd doc\nb\
+          jupyter execute (get-item ccsn\*.ipynb)
+          jupyter execute (get-item presn\*.ipynb)
           jupyter execute AnalyticFluence.ipynb FlavorTransformation.ipynb
           # Download necessary progenitor file, before executing the notebook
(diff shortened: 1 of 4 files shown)


JostMigenda:
The first commit ([50ba643](https://github.com/SNEWS2/snewpy/pull/341/commits/50ba643f3e5779b11de06d6904748a9a2557e2ce)) adds test for Windows[*] without any fixes to the code; to confirm that the tests fail in CI and reproduce the issue observed by our project student.

The second commit ([22b66ce](https://github.com/SNEWS2/snewpy/pull/341/commits/22b66cee3c71e8878d7e2e48f7fdb418bd6d93fa)) adds a fix for the `Warren_2020` model download, as described in #340. This fixes the Warren_2020 test failures, as expected.

Now the only remaining test failure is in `test_flux_container.py:115` (`test_save_and_load`), which has a hard-coded file path that doesn’t appear to work under Windows:
` FileNotFoundError: [Errno 2] No such file or directory: '/tmp/flux.npz'`
I’ll check if there’s an easy fix for this.


[*] and MacOS, for good measure; though that is a UNIX and shouldn’t have as many peculiarities

JostMigenda:
… and commit [4bb76d5](https://github.com/SNEWS2/snewpy/pull/341/commits/4bb76d5fafa0b302741d3bbd34b2a6ebbe3ee6d7) replaces the hardcoded path by using the [`tempfile`](https://docs.python.org/3/library/tempfile.html) module; so it now works on Windows. This is now ready for review.

JostMigenda:
Hm … one remaining issue in the integration tests: The command `jupyter execute ccsn/*.ipynb` fails on Windows ([see lines 136–156](https://github.com/SNEWS2/snewpy/actions/runs/9371073273/job/25799524125#step:7:137) of the log) (and same for presn), but the return code is still successful.

I see a few options:
1. Figure out how to implement that wildcard correctly in a Windows shell
2. Skip that step on Windows
3. Ignore; the other Jupyter notebooks in the step are running successfully, so this is still a useful test

Option 1 would be the right thing to do; but since I don’t have access to a Windows machine, that doesn’t sound like a rabbit hole I particularly want to go down … 🫣

JostMigenda:
After some experimentation, it turns out that PowerShell apparently does not expand wildcard characters (which is not obvious [from the documentation](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_wildcards?view=powershell-7.4) 🤬). Commit [c713c78](https://github.com/SNEWS2/snewpy/pull/341/commits/c713c78a25ef82d06fcea37cba7e0458c8604494) adds the additional code required for this. With that, afaict the test on Windows are equivalent to those on Linux/MacOS and this PR is ready for review.

Sheshuk (approved):
Looks good!

PR #345, opened 2024-06-17
Drop Python 3.8 and 3.9 Support

JostMigenda:
This PR drops support for Python 3.8 and cleans up some backwards compatibility code.

Python 3.8 is [scheduled to reach end-of-life this autumn](https://peps.python.org/pep-0569/#lifespan); and many major packages in the scientific Python ecosystem already started dropping 3.8 support [since April 2023](https://numpy.org/neps/nep-0029-deprecation_policy.html).

@Sheshuk: If there are any places in the SNEWPY 2.0 code where support for Python 3.9 is holding us back, we can consider dropping 3.9 support as well. (NumPy and others started doing that in April 2024.)

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -24,7 +24,7 @@ jobs:
     strategy:
       # Add a list of python versions we want to use for testing.
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.9', '3.10', '3.11', '3.12']
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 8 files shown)


JostMigenda:
Note: The integration tests are currently expected to fail on the `release_v2.0` branch, because of ongoing work on the flavour implementation. That should not stop us from reviewing and merging this PR.

Sheshuk:
> @Sheshuk: If there are any places in the SNEWPY 2.0 code where support for Python 3.9 is holding us back, we can consider dropping 3.9 support as well. (NumPy and others started doing that in April 2024.)

Yes, let's do that! 3.10 introduces easier syntax for type annotations (which we should use more), and [pattern matching](https://docs.python.org/3/reference/compound_stmts.html#match) which is a clear and nice alternative to many if-else checks. It will be nice to use those features in our code.

JostMigenda:
Done and ready for review!

Sheshuk (approved):
Jost, thanks a lot for this work! Everything looks good to me.

PR #350, opened 2024-08-01
Sheshuk/update flavor matrix interface

Sheshuk:
This is a first part of changes, needed for #344 

Expanding the interface of `FlavorScheme` and `FlavorMatrix` with useful methods and syntactic sugar. The goal is to make the code using these classes more expressive and look more similar to the formulas in the papers.
## Example
This allows us to access and define individual matrix elements as well as whole submatrices with a clear interface (example from #344 defining the elements of matrix in `mass_basis<<flavor_basis` space)
https://github.com/SNEWS2/snewpy/blob/818150348d8258ae0978434c644a0faf25069187/python/snewpy/neutrino.py#L178-L193

## Changes in `FlavorScheme`

1. Allowing to use shorter names for neutrinos, omitting `NU_` prefix. Also ignoring the case.
2. Adding the `'NU'` and `'NU_BAR'` names to access the list of all neutrinos or antineutrinos

## Changes in `FlavorMatrix`
1. Useful methods for complex matrix: `real`,`imag`,`abs`, `abs2`, `conjugate`
2. In the `__matmul__` method: added check if the flavor schemes of the matrices' dimensions are compatible. 
3. Adding the `extra_dims` options to constructors `FlavorMatrix.eye` and `FlavorMatrix.zeros` - to initialize arrays of size `(N x M x extra_dims)`. These extra dimensions can be energy and/or time dependency of the transformation
4. Added conversion methods: `convert_to_flavor` method, and also `<<` and `>>` operators, which allows to do the flavor bases conversion in one line, for example `M_3f = (ThreeFlavor<<M<<ThreeFlavor)`
5. Using `np.einsum` to perform the correct matrix multiplication of the matrices and matrix to flux
6. When creating conversion matrix, for example`(ThreeFlavor<<FourFlavor)`, it will check that the matrix is non-zero - to prevent silent zeroing out when there was an intrinsic conversion to an incompatible flavor scheme

See the changes in `test/test_flavors.py` for the demonstration of some new functionality

### DIFF
--- a/python/snewpy/flavor.py
+++ b/python/snewpy/flavor.py
@@ -131,6 +131,10 @@ class FlavorMatrix:
     @property
     def flavor(self):
         return self.flavor_out
+    @property
+    def T(self):
+        "transposed version of the matrix: reverse the flavor dimensions"
+        return FlavorMatrix(array = self.swapaxes(0,1), flavor = self.flavor_in, from_flavor=self.flavor_out)
         
     @classmethod
     def eye(cls, flavor:FlavorScheme, from_flavor:FlavorScheme = None):
@@ -138,7 +142,12 @@ class FlavorMatrix:
         shape = (len(from_flavor), len(flavor))
         data = np.eye(*shape)
         return cls(data, flavor, from_flavor)
-
+    @classmethod
+    def zeros(cls, flavor:FlavorScheme, from_flavor:FlavorScheme = None):
+        from_flavor = from_flavor or flavor
+        shape = (len(from_flavor), len(flavor))
+        data = np.zeros(shape)
+        return cls(data, flavor, from_flavor)
     @classmethod
     def from_function(cls, flavor:FlavorScheme, from_flavor:FlavorScheme = None):
         """A decorator for creating the flavor matrix from the given function"""
--- a/python/snewpy/flavor.py
+++ b/python/snewpy/flavor.py
@@ -258,6 +258,9 @@ def conversion_matrix(from_flavor:FlavorScheme, to_flavor:FlavorScheme):
             # convert from more flavors to TwoFlavor
             return 0.5
         return 0.
+    #check if the conversion matrix is not zero
+    if np.allclose(convert.array, 0):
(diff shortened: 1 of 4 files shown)


sybenzvi:
Integration tests are failing because of [this obsolete flavor transformation (line 152 of base.py)](https://github.com/SNEWS2/snewpy/blob/05df0ed4a3c62e4e846c80f2bb2520346e28ffa2/python/snewpy/models/base.py#L152) in the `SupernovaModel` base class.

python/snewpy/flavor.py
```diff
@@ -1,6 +1,7 @@
 import enum
 import numpy as np
 import typing
+import snewpy.utils
 
 class EnumMeta(enum.EnumMeta):
```
sybenzvi, line 6: Not a new change in this PR, but probably this child class should be called `FlavorEnumMeta` to avoid the possibility of an inadvertent name clash later:

```
from enum import EnumMeta       # did we mean to use this?
from flavor import EnumMeta     # or this?
```

sybenzvi (commented):
Suggestion: while we have this PR open, rename the `EnumMeta` child class in `flavor.py` to`FlavorEnumMeta`.

Otherwise this looks pretty good.

sybenzvi:
> Integration tests are failing because of [this obsolete flavor transformation (line 152 of base.py)](https://github.com/SNEWS2/snewpy/blob/05df0ed4a3c62e4e846c80f2bb2520346e28ffa2/python/snewpy/models/base.py#L152) in the `SupernovaModel` base class.

After discussion on Zoom, we agree that the failing integration tests are all due to the obsolete code in `get_transformed_spectra` that will be not be addressed in PR #351. So this should not hold up the merge.

sybenzvi (approved):
I made a cosmetic change of `snewpy.flavor.EnumMeta` to `snewpy.flavor.FlavorEnumMeta`, which feels a bit safer. Otherwise we can commit this, with the understanding that integration tests will fail in `release_2.0`.

PR #353, opened 2024-08-13
Delete old deprecations and unused code

### LINKED ISSUE #225: `init_model` convenience function

JostMigenda:
Removes a lot of deprecated code from the early days of snewpy v1.x and closes #225.

This should all be uncontroversial stuff. I’ll put the changes to individual `SupernovaModel` child classes into a separate PR later today, since there may be a few questions to discuss there.


JostMigenda:
Integration test failures are due to `get_transformed_spectra` still using the old transformations; this same failure is also present on the `release_v2.0` branch, so shouldn’t stop us from reviewing/merging this PR.

sybenzvi (approved):
No objections, this looks like a straightforward deletion of obsolete warnings and features.

PR #355, opened 2024-08-13
Rate table & snowglobes interface update

evamick:
To the rate table code, added KM3NeT, integrated DarkSide-20k, XENONnT, LZ, and PandaX-4T to be calculated like the rest of the detectors, and fill in spots that were previously "-". As well as formatting it to output formatted to create an html table utilizing mathjax. Also had to add LZ & PandaX-4T into the snowglobes_interface.py for those integrations to work. I added the html rate table code as its own file rather than replacing the existing one, but that could be changed if desired!

### DIFF
--- a/python/snewpy/scripts/SNEWS2.0_rate_table.py
+++ b/python/snewpy/scripts/SNEWS2.0_rate_table.py
@@ -2,9 +2,22 @@ import numpy as np
 import os
 from snewpy import snowglobes
 
+#Select output format, Mathjax or LaTeX
+selection = False
+while selection == False:
+    outputFormat = input('Enter 1 for Mathjax format\nEnter 0 for LaTeX format\n')
+    if outputFormat == '1':
+        mathjax = True
+        selection = True
+    if outputFormat == '0':
+        mathjax = False
+        selection = True
+    if outputFormat != '1' and outputFormat != '0':
+        print('Please enter 1 or 0.')
+
 home_directory = os.getcwd()
 SNOwGLoBES_path = None  # change to SNOwGLoBES directory if using a custom detector configuration
-SNEWPY_models_base = "C:/path/to/snewpy/models/"  # local directory containing model input files ("SNEWPY_models")
+SNEWPY_models_base = "Your/SNEWPY/Models/Path/Here"  # local directory containing model input files ("SNEWPY_models")
 
 d = 10  # distance of supernova in kpc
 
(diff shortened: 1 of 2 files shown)


snowglobes_interface.py
evamick, line 1: I initially uploaded an updated copy of this file rather than just editing the file to have the updates

evamick:
I haven't done a pull request before, I'm sorry if anything isn't done correctly!

python/snewpy/scripts/SNEWS2.0_rate_table.py
```diff
@@ -1,190 +1,193 @@
...
-
-
-ascii.write(data,Writer=ascii.Latex)
+# +50 in line 143 is for rounding so we have a nice number for the table
+
+dettype='km3net'
+mass=69366 * 3
+data['Mass [kt]'][3] = "~"+str(100*int(0.01*(mass*total_events['s27.0']['AdiabaticMSW_NMO'][dettype+"smeared"]/
+      total_events['s27.0']['AdiabaticMSW_NMO'][dettype+"unsmeared"]))+50)+"*"
+# +50 in line 145 to get 150 as an output, the average between normal (100) & inverted (200)
```
JostMigenda, line 150: Isn’t this the same rounding as for `icecube` a few lines above?

python/snewpy/scripts/SNEWS2.0_rate_table.py
```diff
@@ -1,190 +1,193 @@
...
-data['Type'].append('Xe/any $\\nu$')
-data['Mass [kt]'].append(mass)
-data['Location'].append('China')
-data['\\SI{11.2}{\solarmass}'].append(int(mass*s11_epT))
-data['\\SI{27.0}{\solarmass}'].append(int(mass*s27_epT))
-data['\\SI{40.0}{\solarmass}'].append("-")
-
-
-ascii.write(data,Writer=ascii.Latex)
+# +50 in line 143 is for rounding so we have a nice number for the table
```
JostMigenda, line 144: There’s an old adage: “Don’t comment bad code—rewrite it!”

Your comment definitely helps; but I think it’d be even better to change this to use the built-in `round()` function, and get rid of the puzzling `int(0.01*(… + 50))`.

JostMigenda, line 144: There’s an old adage: “Don’t comment bad code—rewrite it!”

Your comment definitely helps; but I think it’d be even better to change this to use the built-in round() function, and get rid of the puzzling `int(0.01*(… + 50))`.

python/snewpy/scripts/SNEWS2.0_rate_table.py
mcolomermolla: We are not sure why it was there for IceCube, and for IceCube, it is there only inside the denominator (unsmeared), not out of the parenthesis...

JostMigenda:
This is looking quite good now; thanks for your work!

Just one more thing: Since Mathjax and LaTeX are so similar, would it be possible to extend the code to output either format (as selected by the user)? LaTeX tables are handy for copying and pasting into papers, so it’d be great to keep that functionality without extra work by the user.
(I think apart from the HTML tags before/after the Mathjax, the changes would be very minor.)

JostMigenda (approved):
Thanks @evamick for the good work and congratulations on your first contribution to snewpy! 🎉

PR #356, opened 2024-08-14
Remove deprecated arguments from SN model initialisers

JostMigenda:
This PR is intended to remove deprecations in the SN model initialisers, including initialising from a file name and inconsistent keyword arguments.

There are still a couple of `@deprecated` decorators we should remove before merging this. (E.g. for many models, the `eos` keyword is marked as deprecated, because there is only a single possible value; so it shouldn’t be required as part of the `__init__`.)
I think the best way to proceed there would be to remove `eos` from the parameter list and manually set `self.metadata["eos"]` in the `__init__` to the fixed value instead. I’ll do that tomorrow and mark this as a draft PR until then; but wanted to get it out now before I go to bed, in case you want to review this during your daytime, @Sheshuk. (Yay for timezones! 😉)

(Also: Yes, I’ve currently marked the branch from #353 as a base for this PR for simplicity; but can rebase it onto the `release_v2.0` branch tomorrow.)

### DIFF
--- a/python/snewpy/models/ccsn.py
+++ b/python/snewpy/models/ccsn.py
@@ -38,7 +38,6 @@ from snewpy.models import ccsn_loaders as loaders
 from .base import PinchedModel
 
 from snewpy.models.registry_model import RegistryModel, Parameter
-from snewpy.models.registry_model import deprecated
 from snewpy.models.registry_model import all_models
 from textwrap import dedent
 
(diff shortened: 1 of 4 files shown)


Sheshuk (commented):
Changes look good so far, and we're heading towards a cleaner and more beautiful future interface!
Looks like you've got everything, concerning with `legacy_filename_initialization`.

I agree with the plan to remove the deprecated parameters and putting them directly into the metadata.

JostMigenda:
Ready for review now!

JostMigenda:
Okay, now the integration tests are back to failing in `get_transformed_spectra` (where we expect them to fail until the flavour transformations are updated to three flavours) and it’s _actually_ ready to review.

sybenzvi (approved):
Changes are mostly removal of unused packages, options, and legacy interfaces. Looks good.

PR #358, opened 2024-08-14
 Sybenzvi/suppress div by zero

sybenzvi:
Copying div-by-zero/invalid suppression to release 2.0.


sybenzvi:
Decision: deleting this without merging, since at some point we'll need to sync `main` with `release_v2.0` anyway and the commit will make its way in then.

PR #361, opened 2024-09-19
fix bugs in README sample code

JostMigenda:


README.md
```diff
@@ -47,11 +47,13 @@ msw_nmo = AdiabaticMSW(mh=MassHierarchy.NORMAL)
 
 … letting you quickly calculate the neutrino flux reaching Earth:
 ```Python
+times = [0.5, 1] * u.s
+energies = range(5,50) * u.MeV
```
Sheshuk, line 51: In our examples in similar context we usually use numpy arrays and `<<` operator:
```suggestion
times = [0.5, 1] << u.s
energies = np.arange(5, 50) << u.MeV
```
But probably it makes sense to avoid it here for maximum simplicity

Sheshuk (approved):
Looks good to me

PR #362, opened 2024-10-15
Support Python 3.13 (and drop 3.8)

JostMigenda:
Python 3.13 was released last week and 3.8 [reached end-of-life](https://devguide.python.org/versions/) at the same time. This PR updates our testing matrix accordingly and removes some compatibility workarounds that are no longer necessary.


Sheshuk (approved):
Looks good!

PR #364, opened 2024-11-14
Add NuFIT 6.0 and PDG 2024 oscillation parameter presets

JostMigenda:
The regular-ish update. This PR leaves NuFIT 5.0 as the default value; though we should probably update that later for the v2.0 release.


Sheshuk (approved):
LGTM, thanks for this update!

PR #369, opened 2024-12-04
Update publish.yml

JostMigenda:
Re-export the name of the published git tag exactly (including the `v` prefix!) in the `get_version` step. This ensures that the tag/title of the GitHub release created in a later step matches the git tag.

(Noticed this while responding to #368; see also discussion there.)


Sheshuk (approved):
Thanks for fixing this.
So this also ensures that there is always a `v` prefix in tags?
Probably it's OK, since we're following this scheme anyway.
Looks good to me

PR #374, opened 2025-02-21
Include notebooks in documentation

### LINKED ISSUE #185: Include notebooks in documentation

JostMigenda:
Closes #185.

~Currently marked as a draft, because I still need to check whether this works with GitHub Actions and Readthedocs.~

----

Preview screenshots (generated locally):

image
image


JostMigenda:
Discovered a couple of minor stumbling blocks while getting this to build on Read The Docs, but everything should be ready now. You can view the [rendered version of the docs for this branch](https://snewpy.readthedocs.io/en/jostmigenda-notebooksindocs/notebooks.html) here.

mcolomerm:
I have a couple of comments:

1) For the snewpy.snowglobes Usage Example, why don't we document it with the newer way? :

`model = Bollig_2016(progenitor_mass=27<<u.Msun)

times    = np.linspace(0,10,10000)<<u.second;
energies = np.linspace(0.25,200.25,400)<<u.MeV

flux = model.get_flux(t = times, E = energies, distance=10<<u.kpc, flavor_xform=AdiabaticMSW(mh=MassHierarchy.NORMAL))

from snewpy.flux import Container

detector='wc100kt30prct_he'
rc = RateCalculator()
rates = rc.run(flux, detector) 

nevents = np.ones((10000,400))
time_list = []
events = np.zeros(10000)
for ch,r in rates.items():
    rT = r.sum('energy')
    for i in range(len(rT.time)):
        time_list[i] = round(rT.time.value[i],3)
        events[i] += rT.array.squeeze().value[i]/1000.
plt.plot(time_list, events)
plt.ylabel(f'number of events per ms bin', size=16)#{rT.__class__.__name__},  {r.array.unit}')
plt.xlabel(f'time [s]',size=16)
`

2) For the SNOwGLoBES Model Access:

I find a bit weird that you call it SNOwGLoBEs model, when SNOwGLoBEs is not related to ccsn models, and also the notebook does not make usage of any snowglobes-related (it is just fluence), no cross-section or anything.

JostMigenda:
Thanks for the comments, @mcolomerm!

1. Note that this newer interface is [still documented as unstable](https://snewpy.readthedocs.io/en/jostmigenda-notebooksindocs/flux.html#:~:text=Note,warning) and not for general usage. That should change with snewpy 2.x, but for now I would leave this as is.
2. The name comes from the fact that this is the format for input fluxes used by SNOwGLoBES. I agree it might be a little confusing; but changing that would require much larger modifications to the underlying code, not just the notebook itself, so it’s outside the scope of this PR. Do you think it would be useful to add 1–2 explanatory sentences to the start of the notebook?

JostMigenda:
Updated the notebooks introduction with a better name and description. I’ve also fixed an outdated import and a bug in a plotting function, which used a hardcoded time instead of the value passed in as an argument.

PR #375, opened 2025-02-23
Add support for late-time 3D models based on Fornax

sybenzvi:
This PR adds software support for new 3D models posted in https://www.astro.princeton.edu/~burrows/nu-emissions.3d.update/.

The model files have been uploaded to the [snewpy-models-ccsn](https://github.com/SNEWS2/snewpy-models-ccsn) repository. Complete the following activities before merging this PR:

- [x] Add unit tests of the new `Fornax_2024` class and models.
- [x] Add a documentation notebook demonstrating the output.
- [x] Make a new tag/release of the [snewpy-models-ccsn](https://github.com/SNEWS2/snewpy-models-ccsn) project.
- [x] Update `snewpy/models/model_files.yml` so that all CCSN models point to the new release repo.


python/snewpy/models/ccsn.py
```diff
@@ -429,6 +430,31 @@ def __init__(self, progenitor_mass:u.Quantity):
...
 
 
+_fornax_2024_progenitors = [
+                'u8.1',  '9a',    '9b',     '9.25',  '9.5',
+                'z9.6',  '11',    '12.25',  '14',    '15.01',
+                '16.5',  '16',    '17',     '18',    '18.5',
+                '19',    '19.56', '20',     '21.68', '23',
+                '24',    '25',    '40',     '60',    '100']
+
+_fornax_2024_masses = [float(re.sub('[A-Za-z]', '', p)) for p in _fornax_2024_progenitors] << u.Msun
```
JostMigenda, line 440: This will map both `9a` and `9b` progenitors to `9.0 * u.Msun`; so the latter would override the former when creating the `_mass_to_progenitor` dict later and we wouldn’t be able to access the `9a` progenitor.

The two both appear to be based on the 9 Msol progenitor from Sukhbold _et al._ 2016; with the caption of fig. 7 in [arXiv:2401.06840](https://arxiv.org/pdf/2401.06840) explaining:
> The only difference between models 9(a) and 9(b) is the imposition of slight velocity perturbations upon infall in the former (Wang & Burrows 2023a).

Since [none of the other progenitors have these velocity perturbations](https://iopscience.iop.org/article/10.3847/1538-4357/ace7b2#apjace7b2s2), do you think we should just drop the `9a` progenitor and document that 9 Msol corresponds to `9b`? Or is it worth the small extra effort to support both?

sybenzvi: A late answer to this question: I think it's OK to drop the 9a progenitor, justified by the fact that the velocity perturbations are not present in any other model.

Your proposal is just to remove `'9a'` from the progenitors list?

sybenzvi:
The `Fornax_2024` models need to be added to the registry and model tests before this can be merged.

sybenzvi:
PR should comply with the discussion in the 2026-06-09 signal prediction telecon, so I'm merging without further review.

PR #379, opened 2025-04-01
Initializing all preSN models with negative time

Sheshuk:
Our supernova models have `SupernovaModel.time` attribute which contains the intrinsic time values - which in many cases define the valid time range of the given model.

For the presupernova models this time range should be negative (i.e. time before collapse).
But currently `.time` is negative only for Odrzywolek, and positive for all others.

This PR corrects that.

### DIFF
--- a/python/snewpy/models/presn_loaders.py
+++ b/python/snewpy/models/presn_loaders.py
@@ -97,7 +97,7 @@ class Patton_2017(SupernovaModel):
         self.interpolated = _interp_TE(
             times, energies, self.array, ax_t=1, ax_e=2
         )
-        super().__init__(times << u.hour, self.metadata)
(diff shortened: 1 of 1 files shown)


JostMigenda (approved):
Good catch, thanks!

I was a bit surprised at first that this has no impact at all on the output of the demo notebooks; but upon closer look, I noticed that we aren’t accessing `model.time` there at all. Instead, we’re passing in negative times, which are converted to positive values in `SupernovaModel.get_initial_spectra` (since logarithmic interpolation of negative time values would break). And that’s also why it makes sense to change the sign only in the `super().__init__` call and not immediately when reading in the values from a file.
_(Just thinking out loud; no response necessary.)_

PR #380, opened 2025-04-04
Extended cooling tail cleanup

JostMigenda:
Follow-up to #312.

* ignore any user-submitted times that overlap with model-provided times
* change default of normalization to match model-provided value at the final time step; this ensures a smooth transition by default, whereas having a fixed default normalization will give unphysical cut-offs as in the screenshot below
* remove `get_integrated_luminosity` function, [as discussed previously](https://github.com/SNEWS2/snewpy/pull/312#discussion_r1768883416)

image


mcolomerm (commented):
I was just wondering: do we want to force this and make the change from the end of the model to the start of the extension abrupt as it is now, instead of it being smooth as before? Otherwise, I don't see any issue to merge this PR.

JostMigenda:
I think there’s a misunderstanding here? The screenshot above shows the abrupt jump that’s already present in the current `main` branch. (That jump was mistakenly introduced during the review of the initial PR, when Segev and I picked the wrong default value.)

_This PR_ fixes that to create a smooth transition by default, as was the original intention.

JostMigenda:
Yep; it now looks like this:
image

mcolomerm (approved):

PR #385, opened 2025-05-12
Split `ThreeFlavorDecoherence`

### LINKED ISSUE #384: Split `ThreeFlavorDecoherence` into “black box” and `in_sn` transformations
> > > I noticed that ThreeFlavorDecoherence is described as Phenomenological transformation and placed in `__init__.py`. It really should be a SN transformation. It's true that Earth matter can't change the decoherence but vacuum transformations could e.g. one of the mass states decays.
> > 
> > My impression was that `NoTransformation`, `CompleteExchange` and `ThreeFlavorDecoherence` are typically used as “black box transformations”—where researchers are interested in some other aspect and use these extreme scenarios to get a rough idea of how flavor transformations could affect studies of that other aspect. So it makes sense to me to keep these three separate?
> > 
> > However, if you think that any users might want to build a `TransformationChain` out of `ThreeFlavorDecoherence` in the SN plus additional transformations in vacuum/earth, then we could turn TFD into a SNTransformation. In that case, would `TransformationChain(in_sn=ThreeFlavorDecoherence())` automatically result in the same flavour equilibration that we get from the current implementation of TFD? If not, that might be very confusing.
> 
> The use case you describe for TFD is a useful one so my suggestion is to have a Phenomenological Transformation in __init__.py called e.g. **Equilibrate** which implements this case, and to have ThreeFlavorDecoherence as prescription in in_sn.py.  The ThreeFlavorDecoherence is useful for describing the effect of strong turbulence in supernovae; the TwoFlavorDecoherence describes the effect of weak turbulence.

_Originally posted by @jpkneller in https://github.com/SNEWS2/snewpy/issues/351#issuecomment-2873095379_

JostMigenda:
Changes by @jpkneller to close #384. Should be merged after #351.

My review comments to follow in a bit …


JostMigenda:
- [ ] Unit tests and integration tests are currently failing due to these changes
- [ ] With the third commit, we now have `ThreeFlavorDecoherence` (defined as `TransformationChain(in_sn.ThreeFlavorDecoherence())`) _and_ `Equilibrate` in `__init__.py`. What (if anything) is the difference between them?
- [ ] What about `TwoFlavorDecoherence`? Do we have the same situation there and would it make sense to split that up into a “black box” transformation and a separate `in_sn` transformation as well?

Sheshuk:
I like the idea of this splitting! But as a user I would find a class name `Equilibrate` misleading: it's a verb, while all other classes are nouns

jpkneller:
> Unit tests and integration tests are currently failing due to these changes

I missed some consequences of the changes. It looks like the dictionaries in snowglobes.py are having problems. 

>  With the third commit, we now have `ThreeFlavorDecoherence` (defined as `TransformationChain(in_sn.ThreeFlavorDecoherence())`) _and_ `Equilibrate` in `__init__.py`. What (if anything) is the difference between them?

It's subtle. Equilibrate != ThreeFlavorDecoherence if in_vacuum != None in a TransformationChain object. Both are useful. 
 
>  What about `TwoFlavorDecoherence`? Do we have the same situation there and would it make sense to split that up into a “black box” transformation and a separate `in_sn` transformation as well?

I don't think a 'black box' version of TwoFlavorDecoherence adds anything useful. It is almost the same as the CompleteExchange prescription.

jpkneller:
> I like the idea of this splitting! But as a user I would find a class name `Equilibrate` misleading: it's a verb, while all other classes are nouns

Let's change it to Equilibration or FlavorEquilibration then.

jpkneller:
> > I like the idea of this splitting! But as a user I would find a class name `Equilibrate` misleading: it's a verb, while all other classes are nouns
> 
> Let's change it to Equilibration or FlavorEquilibration then.

I changed i to FlavorEquilibration

PR #387, opened 2025-05-13
Update flavor.py

jpkneller:
Add functions to convert neutrino names to equivalent strings used in Marley and the PDG particle numbering code


python/snewpy/flavor.py
```diff
@@ -58,6 +58,41 @@ def to_tex(self):
...
+        elif self.is_muon:
+            name += "u"
+        elif self.is_tauon:
+            name += "t"
+        else: 
+            name += "s"            
+        if self.is_antineutrino:
+            name += "bar"
+        return name    
+
```
JostMigenda, line 79: Probably better to avoid having this in core snewpy and move it to the script that transforms snewpy fluences into MARLEY’s JSON format; to avoid bugs if MARLEY changes down the road.

PR #393, opened 2025-05-31
Restore _init_model function.

sybenzvi:
I'd like to restore the `_init_model` function in `python/snewpy/models/__init__.py`, which was removed without a deprecation warning first placed in SNEWPY v1.6. It turns out that ASTERIA uses it.

### DIFF
--- a/python/snewpy/models/__init__.py
+++ b/python/snewpy/models/__init__.py
@@ -1,6 +1,6 @@
 from warnings import warn
 
-from . import ccsn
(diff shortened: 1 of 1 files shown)


sybenzvi:
I am deleting this PR without a review. It turns out that I was too hasty. ASTERIA's problem with the change to `_init_model` was internal, and I was able to make a workaround. No need to restore the `_init_model` interface to snewpy.

Issue #372, opened 2024-12-11
Update oscillation parameter presets


JostMigenda:
In v2.0, we should update the default value of oscillation parameters, see #364. I have no particular preference whether to use NuFIT 6.0 or PDG 2024, but would probably err on the side of NuFIT 6.0, since our current default is an older NuFIT version.

PR #397, opened 2025-06-18
Default oscillation parameters changed to NuFit6.0

mcolomerm:
This should close issue #372

### DIFF
--- a/python/snewpy/neutrino.py
+++ b/python/snewpy/neutrino.py
@@ -483,7 +483,7 @@ parameter_presets = {
 }
    
 
-def MixingParameters(mass_order:MassHierarchy|str='NORMAL', version:str='NuFIT5.0'):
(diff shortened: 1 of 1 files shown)


JostMigenda:
Thanks @mcolomerm!

Could you also update the documentation accordingly, please? Right now, the `ThreeFlavorMixingParameters` class’ docstring still refers to NuFIT5.0; this should be updated (and probably moved, since the default is not set in the `ThreeFlavorMixingParameters` class but rather in the `MixingParameters` function later.)

mcolomerm:
> Thanks @mcolomerm!
> 
> Could you also update the documentation accordingly, please? Right now, the `ThreeFlavorMixingParameters` class’ docstring still refers to NuFIT5.0; this should be updated (and probably moved, since the default is not set in the `ThreeFlavorMixingParameters` class but rather in the `MixingParameters` function later.)

Right, I forgot that. Done now.

Issue #376, opened 2025-03-11
Fornax_2019 angle-dependent fluxes no longer function


sybenzvi:
I discovered several problems with the `Fornax_2019` class, which includes angle-dependent fluxes.

1. The class caches its calculation of the angle-dependent flux, based on a HEALPix map, using an option `cache_flux` in the constructor. The default value is `False` and the class decorator does not seem to allow overriding the default value.
2. The `get_flux` member inherited from `SupernovaModel` cannot handle angle dependence.
3. The innards of the class need to completely upgraded to handle the `FluxContainer` added to snewpy. Additional modernization may be needed.

This class needs a lot of TLC to be brought up to snuff. Do we remove it?

JostMigenda:
> 1. The class caches its calculation of the angle-dependent flux, based on a HEALPix map, using an option `cache_flux` in the constructor. The default value is `False` and the class decorator does not seem to allow overriding the default value.

This is solely a performance optimisation, right? If so, I think it’s not critical to keep this, if this is nontrivial to implement with the decorator. (And who knows, maybe rewriting with the `FluxContainer` would have some performance benefits anyway, so the need for caching decreases?)

> 2. The `get_flux` member inherited from `SupernovaModel` cannot handle angle dependence.

See the related discussion in #165. I’m not 100% happy with the solution I suggested there (moving the angle dependence into the `__init__()`), but it should be pretty straightforward to get that to work.

> 3. The innards of the class need to completely upgraded to handle the `FluxContainer` added to snewpy. Additional modernization may be needed.
> 
> This class needs a lot of TLC to be brought up to snuff. Do we remove it?

Hm … in principle, it would be good to support different observer directions in SNEWPY; there are some interesting analyses that could be done with this. I notice that the `Fornax_2024` models also have directional information, though it’s not implemented in #375 yet. Have you had a look at the format of those? Would we be able to reuse much of the implementation between the 2024 and 2019 models?

sybenzvi:
@JostMigenda: you are correct, the FITS cache is solely a performance optimization. We can figure out a better way... or drop it.

The `Fornax_2024` models are available in ASCII files from the Burrows group, not HDF5, so I'll need to repack the data to recycle the code used in `Fornax_2019`, or come up with a better internal representation.

PR #398, opened 2025-06-18
Fix integration tests

JostMigenda:
Integration tests were broken due to the large rewrite of flavour transformations that’s now merged. (And a few minor issues that have snuck in in the meantime while we weren’t checking the integration tests.) This PR attempts to fix the tests.

There are a couple of open issues remaining:
- [x] `simplerate_integrationtest.py`: Number of events has increased by 0.6%
- [x] `test_05_snowglobes.py`: For `ar40kt` and `AdiabaticMSW_NMO`, event count has decreased by 6–7%
- [ ] An array size issue (`ValueError: Data array of shape (6, 201) is inconsistent with any valid shapes`) in many model notebooks (`Bollig_2016`, `Fischer_2020`, `Kuroda_2020`, `Nakazato_2013`, `OConnor_2013`, `OConnor_2015`, `Tamborra_2014`, `Walk_2018`, `Walk_2019`, `Warren_2020` and `Zha_2021`) and in `FlavorTransformation.ipynb`
- [ ] The `Fornax_2019` notebook is broken (see #376)
- [ ] The `Fornax_2021`/`Fornax_2022` notebooks have some array size issue in plotting code
- [ ] `AnalyticFluence.ipynb`: fails because `generate_fluence()` hardcodes the `ccsn_loaders` module and `Analytic3Species` doesn’t have a separate loader class


JostMigenda:
For the `simplerate_integrationtest.py`, the difference is significant, but very channel-dependent. I figured it out eventually, but was already most of the way through writing up my investigation, so you get to read that anyway. Enjoy!

```
# this branch
ibd       : 12577.666 events
nue_O16   : 84.792 events
nuebar_O16: 161.729 events
nc        : 275.397 events
e         : 1023.290 events
Total events in Super-K-like detector: 4519.31979825849

# snewpy v1.6
ibd       : 12506.335 events
nue_O16   : 110.435 events
nuebar_O16: 137.741 events
nc        : 275.397 events
e         : 1007.176 events
Total events in Super-K-like detector: 4491.866729366276
```
Notably, `nc` remains unchanged; which strongly hints that the issue is related to the flavour transformation.
This is further confirmed by checking numbers with `NoTransformation`, which are effectively identical for both cases (<0.02% difference in electron scattering channel).

For `AdiabaticMSW_IMO`, both give the same result … namely, exactly the 4519 events that this branch also gives for the NMO case. 🤔

After some experimenting, I found this minimal reproduction of the issue:
```Python
from snewpy.flavor_transformation import AdiabaticMSW
from snewpy.neutrino import MixingParameters

NMO = MixingParameters('NORMAL')
IMO = MixingParameters('INVERTED')

nmo = AdiabaticMSW(NMO)
nmo.P_ff(0, 0)[0,0]
# prints out 0.02220624848915225

imo = AdiabaticMSW(IMO)

# initialising `imo` changes somehow affects `nmo`?!?
nmo.P_ff(0, 0)[0,0]
# prints out 0.29703752184248233
```

Alec and I then dug through the code to investigate how this happens. We finally traced it back to the `ThreeFlavorTransformation` class, where `_mixing_params` was defined as a class attribute, shared between all instances of the class. (I.e., `nmo` and `imo` in the above code example.) Since the setter doesn’t _set_ `self._mixing_params` but just calls a method on it, all instances still share the same value.

Toy example:
```Python
class Foo():
    x = []

f = Foo()
g = Foo()
f.x is g.x  # True

# calling a method on the existing list
f.x.append(1)
f.x is g.x  # still True, now f.x and g.x are both [1]

# reassigning a new value to f.x
f.x = [1, 2]
f.x is g.x  # now False, f.x is [1, 2] while g.x remains [1]
```

This is fixed by the commit I just added; which initialises `_mixing_params` as an instance attribute.

Sheshuk (approved):
Thanks for fixing my bug with static `mixing_pars`!
I confirm that the integration tests are passing. The changes look good to me.

PR #400, opened 2025-06-20
Remove support for `$SNOWGLOBES` environment variable

### LINKED ISSUE #371: remove support for `$SNOWGLOBES` variable to use custom SNOwGLoBES directory
> as we're using `snowglobes_data`, we can drop the support of `$SNOWGLOBES` variable. If a user needs to use a custom snowglobes, they can use `pip install $SNOWGLOBES/snowglobes_data` to use their package in snewpy.

_Originally posted by @Sheshuk in https://github.com/SNEWS2/snewpy/issues/370#issuecomment-2532597258_

JostMigenda:
Closes #371; see reasoning there.

### DIFF
--- a/doc/source/contributing.rst
+++ b/doc/source/contributing.rst
@@ -42,15 +42,11 @@ You can then run the tests using the
 
     pytest
 
-command in the SNEWPY root directory. To skip integration tests which depend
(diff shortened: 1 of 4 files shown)


Sheshuk (approved):
Looks good!

PR #401, opened 2025-07-03
Initializing the mixing_parameters in the c-tors

Sheshuk:
There was a problem in `Three/FourFlavorTransformation` class - the `mixing_params` used to be a static (class) variable. That means that creating two instances of the same class with different mixing parameters, those will be overwritten:
```python
import snewpy.flavor_transformation as flx
from snewpy.neutrino import MixingParameters

msw_NMO = flx.AdiabaticMSW(MixingParameters('NORMAL'))
msw_IMO = flx.AdiabaticMSW(MixingParameters('INVERTED'))
print(id(msw_NMO.transforms.in_sn.mixing_params) == id(msw_IMO.transforms.in_sn.mixing_params))
#True - before this PR
```

I fix it by adding explicit constructor `__init__(mixing_pars)` which initializes the parameters.

The downside of this is that, as always, any subclass of `Three/FourFlavorTransformation` must call `super().__init__()`.

### DIFF
--- a/python/snewpy/flavor_transformation/TransformationChain.py
+++ b/python/snewpy/flavor_transformation/TransformationChain.py
@@ -25,8 +25,8 @@ class TransformationChain(FlavorTransformation):
     """
     def __init__(self,
                  in_sn: SNTransformation,
-                 in_vacuum: VacuumTransformation=NoVacuumTransformation(),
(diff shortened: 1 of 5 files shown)


Sheshuk:
Oh, I see it's already fixed in #398, so I close this PR as a duplicate

Issue #223, opened 2022-10-12
Make signature for `SupernovaModel.get_*_spectra` uniform for all models/subclasses


Sheshuk:
# What we have now
The signature signatures for these functions in the base `SupernovaModel` class are:
https://github.com/SNEWS2/snewpy/blob/b35379952fe18aedc28cdb106f1ee9e7ad08b082/python/snewpy/models/base.py#L99
https://github.com/SNEWS2/snewpy/blob/b35379952fe18aedc28cdb106f1ee9e7ad08b082/python/snewpy/models/base.py#L126

Currently most of the models conform to the base signatures, but some models require additional parameters, like `Fornax*`:
https://github.com/SNEWS2/snewpy/blob/b35379952fe18aedc28cdb106f1ee9e7ad08b082/python/snewpy/models/ccsn.py#L664

## Why is it bad

Having different signatures in base and children classes might lead to confusion and errors. In the Fornax2019 case, for example, there is no way to pass the additional parameters to the `get_transformed_spectra`, so running the following code:
```python
from snewpy.models import ccsn
from snewpy.flavor_transformation import NoTransformation

model = ccsn.Fornax_2019('models/Fornax_2019/lum_spec_10M.h5')
model.get_transformed_spectra(t,E, NoTransformation())
```
results in 
```
TypeError: Fornax_2019.get_initial_spectra() missing 2 required positional arguments: 'theta' and 'phi'
```

# What I suggest

Add the `**kwargs` argument to the signatures:
```python
def get_initial_spectra(self, t, E, flavors=Flavor, **kwargs):
    ...
def get_transformed_spectra(self, t, E, flavor_xform, **kwargs):
    ...
```

## Optional (subjective opinion)

Method `get_initial_spectra` has `flavors` argument, which is not present in `get_transformed_spectra`. This also adds some confusion - as a user I would expect that these methods have the same arguments, apart from the `flavor_xform`.
I see no need to pass flavors to `get_initial_spectra` - if a user needs only specific flavors, he can just filter what he needs from the output. 

So I would even go for 
```python
def get_initial_spectra(self, t, E, **kwargs)
```

JostMigenda:
For `Fornax_2019`, the issue of `theta` and `phi` was already raised in #165. (Though I didn’t get to implement the discussed fix during the hackathon and lost track of it afterwards; thanks for the reminder!)
[...]

PR #402, opened 2025-07-04
Sheshuk/fix model notebooks

### LINKED ISSUE #392: Review & Update the usage examples' notebooks
After the interface changes in preparation of v2.0 we'll need to revisit and update our usage example notebooks.

Sheshuk:
Closes #392 , #223 

Trying to fix the remaining issues with building the notebooks:
 - [x] Errors in most `PinchedModel`-derived model classes, because the `flux.Container` failed to initialize with a scalar time (shape mismatch)
 - [x] Errors in `Fornax_*` models - caused by extra parameters `interpolation`, `phi` and `theta`. We need to discuss how to handle this in #223
 - [ ] Error in `Fornax_2019` model: #403 
 - [ ] Errors in `SNOwGLoBES_models` nb - it needs updating, it uses path to already created tar.gz file.
 - [ ] Errors in `AnalyticFluence` - it's incompatible with our ccsn_loaders.
 
Also, currently building the notebooks in the docs is not very informative: if one fails, we don't get any additional information. I'm adding a separate step, calling the pytest on notebooks, so we can see exctly which ones of them are failed, instead of exit on first failure.


doc/source/conf.py
```diff
@@ -79,6 +79,7 @@
 # directories to ignore when looking for source files.
 exclude_patterns = ['nb/dev']
 
+nbsphinx_allow_errors = True
```
JostMigenda, line 82: I’d like to be very strict here, because any errors that slip through will make it to the published documentation.

Sheshuk: It is still strict: I've added the `pytest --nbval` step, so we're still having the notebooks checked, but it provides more information: instead of just failing on the first failed notebook during the documentation build, it will run each of them as test and give the output.

JostMigenda: @Sheshuk I understand about the additional information by `pytest --nbval`; but I don’t think it’s necessary to allow errors here in `conf.py` to get that additional information, is it?

If _building the docs_ errors out if we have a broken notebook, then it will also error out in readthedocs, so we’re guaranteed never to successfully build broken documentation.
If building the docs doesn’t error out and we have rely on a different step in this CI workflow to alert us of any issues, then there’s always the risk of us overriding that failed test, or modifying this workflow without adjusting the docs build, or … and ending up with documentation that’s broken without us realising it.

JostMigenda (commented):
Sorry for the delay! I didn’t get to this before the conference, then spent most of last week ill with some conference virus … 🤒

Aside from the separate discussion about `nbsphinx_allow_errors` above, I think there’s one more open issue here: In addition to the more exotic model notebooks, which we’ll leave for separate PRs, we also have `FlavorTransformation.ipynb` failing in the integration tests, with the following error:

```
TypeError                                 Traceback (most recent call last)
Cell In[6], line 1
----> 1 fig = plot_total_flux(model, AdiabaticMSW(MixingParameters()), AdiabaticMSW(MixingParameters('INVERTED')))
      2 fig.show()
      3 # fig.savefig('flux_adiabaticmsw.pdf')

Cell In[4], line 37
     35 for flavor in Flavor:
     36     for j, E in enumerate(energies):
---> 37         ispec[flavor][j] /= (4.*np.pi*d**2)
     38         ospec_nmo[flavor][j] /= (4.*np.pi*d**2)
     39         ospec_imo[flavor][j] /= (4.*np.pi*d**2)

TypeError: unsupported operand type(s) for /: 'd2NdEdT' and 'float'
```
This looks like it could be related to the container changes in this PR?

PR #404, opened 2025-07-09
Move model classes from ccsn.py to base.py

jpkneller:
Moved the SNOwGLoBES class and Analytic3Species class from ccsn.py into base.py.


jpkneller:
The Integration Test errors are in the ipynb files which still import Analytic3Species from ccsn.py

jpkneller:
I added just the PISN models for now to see if this will work.

jpkneller:
All the differences between the Type Ia and PISN models with the CCSN models have now been removed. The raw data files for the PISN and Type Ia contain time snapshots and as you noticed, they also contained the unoscillated spectra. By using the NoOsc data it is possible for the user to apply any of the in_sn transformations - the user would not be 'double applying'. It is true that applying some of the flavor prescriptions may not make sense physically but the onus is on the user to know why of the that is.

JostMigenda:
Looking again at this PR, together with #415 and #416.

* As discussed above, I think `Analytic3Species` can stay in the `models.ccsn` module.
* Moving `SNOwGLoBES` to `models.base` is unnecessary if we’re about to replace it with the new `PISN` and `TypeIa` classes. I’ll open an issue to deprecate that class.
* And the update to `python/snewpy/models/model_files.yml` is duplicated in #415.

So overall, let’s close this PR and continue work on the others.

PR #405, opened 2025-07-11
Added Marley Script

### LINKED ISSUE #352: add script that produces output suitable for MARLEY
Request from @schol: add a script that can transform output from snewpy into a JSON format readable by [MARLEY](https://www.marleygen.org/), the DUNE LiAr event generator.

Notes:
1. This will not be a core part of snewpy but we can add a script to the scripts folder.
2. In principle this script might be better off migrating to MARLEY in the future.

jpkneller:
Script to create a tarfile containing MARLEY config files from a particular SN model

### DIFF
--- a/python/snewpy/scripts/snewpy_to_marley.py
+++ b/python/snewpy/scripts/snewpy_to_marley.py
@@ -1,0 +1,95 @@
+
+import logging
+import os
+import re
+import tarfile
+import io
+import json
+
+import numpy as np
+import pandas as pd
+from astropy import units as u
+from warnings import warn
+
+from snewpy.models.ccsn import Bollig_2016
+from snewpy.flavor_transformation import AdiabaticMSW
+from snewpy.neutrino import MixingParameters
+from snewpy.flavor import ThreeFlavor
+
+def marley_name(flavor):
+    marley_name = "v"
+    if 'E' in flavor.name:
+        marley_name += "e"
+    elif 'MU' in flavor.name:
+        marley_name += "u"
+    else:
+        marley_name += "t"
+    if 'BAR' in flavor.name:
+        marley_name += "bar"
+
+    return marley_name  
+    
+def save_as_marley(fluence_object, output_filename, marley_additional_inputs=None):
+    """Save the contents of a fluence object in MARLEY format.
+
+       For each time bin 6 Marley config files are genreated and placed into a tarfile.
+
+       Parameters
+       ----------
+       fluence_object : flux container
+           Input file containing neutrino flux information from supernova model.
+       marley_additional_inputs : dict
+           Additional dictionary that will be inserted into the Marley config files. 
+
+       Returns
+       -------
+       str
+    	   Path of tarfile with the Marley config files.
+       """
+
+    times=fluence.time
+    ntbins = len(times)
+
+    if marley_additional_inputs != None:
+        marley_data = marley_additional_inputs
+    else: 
+        marley_data = {}
+
+    if 'reactions' not in marley_data:
+       marley_data['reactions'] = [ "ve40ArCC_Bhattacharya2009.react", "ES.react" ] 
+       
+    marley_data["source"] = { 'type': "histogram", 'E_bin_lefts': fluence_object.energy[:-1].value.tolist(), 'Emax': fluence_object.energy[-1].value } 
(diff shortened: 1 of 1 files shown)


python/snewpy/scripts/snewpy_to_marley.py
```diff
@@ -0,0 +1,95 @@
...
+
+times    = snmodel.time
+energies = np.linspace(0,50,501)<<u.MeV
+
+d = (10*u.kpc).to('cm').value # distance to SN
+    
+mp_nmo = MixingParameters()
+xform = AdiabaticMSW(mp_nmo)
+
+flux = snmodel.get_flux(t=times, E=energies,  distance=d, flavor_xform=xform)
```
JostMigenda, line 90:
```suggestion
d = 10 * u.kpc  # distance to SN
    
mp_nmo = MixingParameters()
xform = AdiabaticMSW(mp_nmo)

flux = snmodel.get_flux(t=times, E=energies,  distance=d, flavor_xform=xform)
```
The `get_flux` function is able to handle Astropy quantities, so there’s no need to convert to a numerical value. (Additionally, if the `distance` is given as a unit-less number, `get_flux` will [assume that it is in kpc](https://github.com/SNEWS2/snewpy/blob/17f14724e679261f985bc503531b630030ee4172/python/snewpy/models/base.py#L182), so converting to a cm value will give a wrong result.

python/snewpy/scripts/snewpy_to_marley.py
```diff
@@ -0,0 +1,95 @@
...
+import os
+import re
+import tarfile
+import io
+import json
+
+import numpy as np
+import pandas as pd
+from astropy import units as u
+from warnings import warn
```
JostMigenda, line 12:
```suggestion
import re
import tarfile
import io
import json

import numpy as np
from astropy import units as u
```
Remove unused imports.

python/snewpy/scripts/snewpy_to_marley.py
```diff
@@ -0,0 +1,95 @@
...
+        marley_name += "t"
+    if 'BAR' in flavor.name:
+        marley_name += "bar"
+
+    return marley_name  
+    
+def save_as_marley(fluence_object, output_filename, marley_additional_inputs=None):
+    """Save the contents of a fluence object in MARLEY format.
+
+       For each time bin 6 Marley config files are genreated and placed into a tarfile.
```
JostMigenda, line 35:
```suggestion
       For each time bin, 6 Marley config files are generated and placed into a tarfile.
```

python/snewpy/scripts/snewpy_to_marley.py
```diff
@@ -0,0 +1,95 @@
...
+
+       Parameters
+       ----------
+       fluence_object : flux container
+           Input file containing neutrino flux information from supernova model.
+       marley_additional_inputs : dict
```
JostMigenda, line 41: It might be nice to rename this to `additional_inputs`, to be a bit less unwieldy?

python/snewpy/scripts/snewpy_to_marley.py
```diff
@@ -0,0 +1,95 @@
...
+           Additional dictionary that will be inserted into the Marley config files. 
+
+       Returns
+       -------
+       str
+    	   Path of tarfile with the Marley config files.
+       """
+
+    times=fluence.time
```
JostMigenda, line 50: Should this be `fluence_object.time`?

(Actually, it might be nice to rename that function argument to just `fluence`. The `_object` suffix just takes up space but doesn’t add any information—everything is an object in Python.)

JostMigenda (changes requested):
I’ve noticed one subtle issue which could give wrong results (by some orders of magnitude), plus a few other bits of feedback. See the individual comments for details.

By the way, have we heard back from the MARLEY user(s) who wanted to test this? It would be good to get their feedback and confirm it works before we merge this.

JostMigenda (approved):
All my comments are resolved.

I’d still prefer to wait for feedback from actual MARLEY users; but if they continue to be unresponsive for the next few weeks, at some point we can just go ahead and merge this as is.

jpkneller:
I implemented the feedback I got from Yun-Tse Tsai and now the script inserts all the pieces into the config files that MARLEY needs.

PR #409, opened 2025-09-25
Update in_earth.py

jpkneller:
Modified P_fm function to add try-except block around allclose check otherwise you get an error message if the lengths of E and prior_E don't match.


JostMigenda (approved):
LGTM, thanks!

PR #373, opened 2024-12-20
replace deprecated usage of setup.py

JostMigenda:
Closes #346.

Python packaging documentation [explicitly warns](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#:~:text=However,run%20anymore) that commands like `python setup.py sdist bdist_wheel` are deprecated and “MUST NOT be run anymore”.
This PR modernizes our publishing workflow as recommended there. The [`build` module](https://build.pypa.io/en/stable/) ensures more reproducible builds by requiring build dependencies to be explicitly declared (in a `pyproject.toml` file) and building the project in an isolated virtual environment. As a nice side effect, this PR also cleans up some helper code which was no longer necessary.

PR #413, opened 2025-10-13
Support Python 3.14 (and drop 3.9)

JostMigenda:
Python 3.14 was released last week and now that `h5py` has [wheels available on PyPI](https://github.com/h5py/h5py/issues/2651#issuecomment-3382030369), let’s add it to our testing matrix!

This PR also drops support for 3.9, which [reaches end-of-life](https://devguide.python.org/versions/) this month. (And if, like me, you’re wondering “Didn’t we drop 3.9 already?” … the answer is yes, [we did](https://github.com/SNEWS2/snewpy/pull/345), back when v2 work was on a separate branch. Then _later,_ #373 touched the old metadata on the `main` branch and “won” in the eventual merge.)


sybenzvi (approved):
After fixing a conflict with `main`, this is ready to be merged.

PR #415, opened 2025-11-05
PISN & type Ia models - On hold

### LINKED ISSUE #434: Deprecate `SNOwGLoBES` model class
In snewpy v1.x, the `models.ccsn.SNOwGLoBES` class was used to access the type Ia and PISN models. Now, #415 and #416 implement dedicated classes for those models (which are more consistent with other snewpy model classes), replacing the `SNOwGLoBES` class.

In principle, there is still one other use case for that class—generating custom fluence in SNOwGLoBES, then using the `SNOwGLoBES` class to read those in for further analysis in snewpy. We briefly mention that in the `SNOwGLoBES_models.ipynb` notebook; however, in practice I think it’s very unlikely that anyone is actually using this: SNOwGLoBES itself comes with a small number of fairly old models that should no longer be used for new analyses; and preparing new model files for SNOwGLoBES is about the same work as preparing them for snewpy. At the same time, snewpy-native models offer much more functionality than the `SNOwGLoBES` class, which e.g. has no way to apply flavour transformations.

Therefore, I’d suggest marking the `SNOwGLoBES` class as deprecated. (If we want to be absolutely certain, we could include a note asking users to open an issue if they use that class in a way that’s not replaced by the dedicated classes.)

jpkneller:
New pisn model classes and Jupyter workbook


JostMigenda:
I’ll try to review this in details tomorrow; here are some initial, high-level thoughts:

* For both models, tests are missing.
* The loader classes for both models look nearly identical; which makes sense given they’re the exact same format. Are there relevant differences or would it make sense to de-duplicate that?
* The `TypeIa` class doesn’t match our `Lastname_YYYY` convention for model classes. Can we rename that to `Wright_2016`? (Which matches the year both preprints appeared on the arXiv; even if the PRD paper on the GCD model is from 2017.)
* I notice the files for PISN and type Ia are in a [combined repo](https://github.com/SNEWS2/snewpy-models-pisn_typeIa/), but the classes are in separate files. I don’t have a strong preference either way; happy to leave it as is, but wanted to flag it in case people see advantages to one or the other. (If joint, maybe `snewpy-models-exotics`/`snewpy.models.exotics` would be a better name, just in case there’s ever any other exotic models.)

python/snewpy/models/pisn_loaders.py
```diff
@@ -0,0 +1,124 @@
...
+        spectra : dict
+            Dictionary of model spectra, keyed by neutrino flavor.
+        """   
+        #convert input arguments to 1D arrays
+        t = u.Quantity(t, ndmin=1)
+        E = u.Quantity(E, ndmin=1)
+
+        initial_spectra = {}
+        for flavor in ThreeFlavor:
+            initial_spectra[flavor] = self.interpolation[flavor]((t, E)) / (u.MeV * u.s)
```
JostMigenda: If both `t` and `E` are multi-element arrays (e.g. if we do `model.get_flux(times, energies, d)` as demonstrated in `FlavorTransformation.ipynb`), this will fail.

The fix could be something like:

```python
        t = u.Quantity(t, ndmin=1).to(u.s).value
        E = u.Quantity(E, ndmin=1).to(u.MeV).value
        tE_grid = np.stack(np.meshgrid(t, E, indexing='ij'), axis=-1)

        initial_spectra = {}
        for flavor in ThreeFlavor:
            initial_spectra[flavor] = self.interpolation[flavor](tE_grid) / (u.MeV * u.s)
```
(Note that we also need to remove the units in between, otherwise `np.stack` will complain that we’re mixing units.)

I’ve played around with a toy example to figure out the correct order of indices (`self.interpolation[flavor](tE_grid)[i, j] == self.interpolation[flavor]((t[i], E[j]))`), but since this is so easy to mix up, I would appreciate if someone could double-check that.

JostMigenda: The same applies to the `TypeIa` class.

jpkneller:
>     * For both models, tests are missing.

I'll fix that.

>     * The loader classes for both models look nearly identical; which makes sense given they’re the exact same format. Are there relevant differences or would it make sense to de-duplicate that?

I recycled the SNOwGLoBES model class as a parent for both. This model class lives on. 
 
>     * The `TypeIa` class doesn’t match our `Lastname_YYYY` convention for model classes. Can we rename that to `Wright_2016`? (Which matches the year both preprints appeared on the arXiv; even if the PRD paper on the GCD model is from 2017.)

I am okay with this suggestion. Do you want to keep TypeIa for backwards compatibility? 

>     * I notice the files for PISN and type Ia are in a [combined repo](https://github.com/SNEWS2/snewpy-models-pisn_typeIa/), but the classes are in separate files. I don’t have a strong preference either way; happy to leave it as is, but wanted to flag it in case people see advantages to one or the other. (If joint, maybe `snewpy-models-exotics`/`snewpy.models.exotics` would be a better name, just in case there’s ever any other exotic models.)

I made a single repo so there would be fewer to manage: I'd like to keep the models in separate files.

JostMigenda:
> > * The `TypeIa` class doesn’t match our `Lastname_YYYY` convention for model classes. Can we rename that to `Wright_2016`? (Which matches the year both preprints appeared on the arXiv; even if the PRD paper on the GCD model is from 2017.)
> 
> I am okay with this suggestion. Do you want to keep TypeIa for backwards compatibility?

The `TypeIa` class only ever appeared on this PR branch and was never included in any snewpy release; so I don’t think there is any compatibility issue if we rename it here

PR #416, opened 2025-11-06
TypeIa SN models

jpkneller:


jpkneller:
The TypeIa.ipynb notebook seems to be having trouble downloading the data from the repo. It downloads a fragment then stops. It is not clear to me why this happens because it's exactly the same code as for the other registry models. If the full dataset is supplied, the notebook works just fine.

JostMigenda:
Since this PR is very similar to #415 and e.g. needs the updates to `model_files.yml` that are already included in that PR, I’ll merge this into that branch and review both of them at once (probably this afternoon).

PR #419, opened 2026-02-02
Ash/fix earth matter neutrino basis

Sheshuk:
fixing the interface to BEMEWS:
* The flavor/mass bases as flipped
* The matrix dimension added for the time (to ensure correct matrix multiplications)


JostMigenda (approved):
LGTM, thanks!

PR #421, opened 2026-03-26
Fornax_2019 fix #420

adamnigel00:
Fix Fornax_2019 array time input, unit conversion, and Python 3.10+ import error

- flavor.py: add `from __future__ import annotations` to defer evaluation
  of `FlavorScheme | None` type annotations, which caused a TypeError at
  import time on Python 3.10+ due to FlavorEnumMeta.__or__ not supporting
  NoneType (Issue 1)

- flavor_xform=AdiabaticMSW(mh=MassHierarchy.NORMAL) ->
flavor_xform=AdiabaticMSW(MixingParameters(MassHierarchy.NORMAL)) (Issue 2) (for converting flux to events)

- ccsn_loaders.py: fix Fornax_2019._get_initial_spectra_dict to handle
  array time input correctly. Previously _get_binnedspectra was called with
  the full time array instead of a scalar, causing either a broadcast
  ValueError or silently wrong results (all time steps mapped to the same
  model index). Now finds nearest time indices upfront, uses bulk HDF5
  reads per (ebin, l, m) across unique time steps, and remaps to the full
  time array via j_inverse (Issue 3a)

- ccsn_loaders.py: fix unit conversion in _get_initial_spectra_dict —
  divide luminosity spectrum [erg/(MeV*s)] by neutrino energy E to obtain
  the number spectrum [1/(MeV*s)] required by flux.Container (Issue 3b)

### DIFF
--- a/python/snewpy/flavor.py
+++ b/python/snewpy/flavor.py
@@ -1,3 +1,5 @@
+from __future__ import annotations  # fix: defer annotation evaluation so `FlavorScheme | None` is not evaluated at class-body parse time (see GitHub issue)
+
 import enum
 import numpy as np
 import typing
--- a/python/snewpy/models/ccsn_loaders.py
+++ b/python/snewpy/models/ccsn_loaders.py
@@ -585,46 +585,81 @@ class Fornax_2019(SupernovaModel):
         """
         initialspectra = {}
 
-        # Extract the binned spectra for the input t, theta, phi:
-        _E, _dE, _spec = self._get_binnedspectra(t, self.theta, self.phi)
-
         # Avoid "division by zero" in retrieval of the spectrum.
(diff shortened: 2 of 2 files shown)


adamnigel00:
@JostMigenda Hi Jost, apologies as this PR was definitely rushed on my end, without going through proper verification so long as it worked and produced agreeable results. I will convert this to a draft as I would like to ensure this commit is reliable before any review.

With regards to flavor_xform, this was intrinsic to a script, and indeed not a part of the diff.

sybenzvi:
I'm going to close this PR because it's been superseded by #417 and #427. The suggestions made here to fix `ccsn_loaders.py` have been partially implemented (some of the suggested updates were incorrect). We are not implementing the type annotations as the problem reported is not reproducible.

PR #426, opened 2026-05-07
Use valid SPDX license expression in pyproject.

sybenzvi:
PR to close out #382.


pyproject.toml
JostMigenda:
```suggestion
requires = ["setuptools>=77"]
```
First `setuptools` version that [supports the new license expression](https://setuptools.pypa.io/en/latest/history.html#v77-0-0)

sybenzvi: Thanks @JostMigenda, I applied your commit.

PR #428, opened 2026-05-11
Use healpy as default; astropy-healpix on Windows.

sybenzvi:
It appears that `astropy-healpix` does not have great support for plotting HEALPix maps, so an option is conditional installation of `healpy` on Linux/Mac OS and `astropy-healpix` on Windows.

If I'm incorrect and have missed some useful plotting utilities, we can alter or delete this branch.


JostMigenda:
Hm … are we using the package(s) to plot HEALPix maps anywhere in snewpy itself?

It looks like both packages have all the functionality we are actually using (which, from a quick search of `ccsn_loaders.py`, does not seem to involve plotting). If that’s the case, I’d prefer to avoid the complexity of having different dependencies on different systems.

Another potential problem with healpy is that they seem to have some delay in supporting new Python releases; e.g. it didn’t support Python 3.14 until ~2 months after its release. In contrast, astropy-healpix uses the [stable ABI](https://docs.python.org/3/c-api/stable.html) (`abi3`), so existing wheels should run e.g. under the upcoming Python 3.15, without requiring a new build & release.

jpkneller:
I am not aware of any use of healpy to make HEALPix maps anywhere in snewpy. I'm okay with sticking with the astropy version for all platforms.

sybenzvi:
We do use `healpy.mollview` in one of the Fornax plotting notebooks, though not in the core code. It doesn't sound like we have a consensus to support both versions, so I'll close this PR without merging, but leave the branch alive for a bit longer.

PR #432, opened 2026-05-29
Jpkneller reorganize doc folder

jpkneller:
Moved all the usage example into /python/examples/ 
Reorganized /doc folder


sybenzvi:
I think the consensus on the signal prediction call (2026-06-09) was that notebooks should live in `docs/nb` and the scripts should be moved from `python/snewpy/scripts` into something like `docs/examples` or `docs/scripts`.

jpkneller:
> I think the consensus on the signal prediction call (2026-06-09) was that notebooks should live in `docs/nb` and the scripts should be moved from `python/snewpy/scripts` into something like `docs/examples` or `docs/scripts`.

Fine with me. I don't have a strong preference for the location as long as the scripts and notebooks are nearer to each other than they are right now.

sybenzvi:
Rather than revert changes to config files and pydocs, I am closing this PR and implementing the agreed-upon changes in PR #435. Please have a look, @jpkneller.

PR #439, opened 2026-06-17
fix url to Myers 2026 models

mckzmyers:
Fixed an incorrect URL to the Myers models in the `snewpy-models-presn` repo.


jpkneller (approved):

JostMigenda:
Huh, good catch!

I’m a little surprised at this; `main` has been the default branch name for years and is also what we use in the main snewpy repo and the other two `snewpy-models-*` repos. So I’d rather leave the URL here as is and update the repo configuration instead; so that inconsistency doesn’t come back to bite us again and again in the future.

JostMigenda:
@Sheshuk just updated the presn models repo, so the existing URL works now. Thanks!

PR #441, opened 2026-06-22
Update snowglobes.py - Experimental

jpkneller:
- Single generate function accepts a model class (not a string) and input times and energies. Returns Container object of flux or fluences. generate_time_series and generate_fluence are retained but are deprecated. 
- New function called calculate combines action of simulate and collate. The input is a flux or fluence Container (i.e. return object from generate) _or_ a string with the name of the file written by generate. Only calculates 'smeared' or 'unsmeared' as given by 'detector_effects' argument, not both. If string is input, a numpy archive file is made and the return is the name of the file: if not then then the returns is a nested dictionary (detector : channel) of event rates / numbers.


jpkneller:
This is a re-write of the functions in snowglobes.py with some additional small changes elsewhere. 
The principle changes are:

1. The two generate_* functions are now a single function. the arguments tmin and tmax are replaced with either a single time or an array of times. If a single time, a flux at that time is returned as a flux.Container object; if an array then the fluence in each time bin is returned as the flux.Container object. It is now also possible to set the limits on the energies. The flux / fluence tables is / are saved in a npz file, either with a name provided by the user or one made from the model name. This required adding a name data member to the model classes. 
2. The arguments to the simulate function have been changed to allow the passing in of the flux.Container object returned by the generate function as well as the name of the npz file. The detector_effects argument is now used - previously it was ignored. If a file was read, the event rates / numbers are saved in another file. The function now returns the event rates or event numbers as a double dictionary of flux.Container objects. The first key is the 'SNOwGLoBES' detector type name, the second the channel in that detector. 
4. The collate function now accepts the double dictionary of flux.Container objects returned by the simulate function as well as the name of a npz file with the event rates / numbers. This function now returns another double dictionary of flux.Containers with aggregated channels. The aggregated channels now include the coherent scattering reactions. If a file was read, a new npz file is written with the collated tables.

jpkneller:
FYI this new version is able to reproduce the table on the SNEWS website and within 1% of the numbers in _rate_crosscheck_table.py

JostMigenda:
Hi Jim! I’m on annual leave for much of this week and will be attending conferences the following 1.5 weeks, so may not be able to review this in great detail before the end of the month. That said, it looks like you’re still working on it actively (with about 30 commits since you requested the review); so please let me know when it settles down and is ready for a thorough review.

In the meantime, my first, high-level thoughts are:
* We haven’t marked the `generate_*` functions as deprecated in any earlier snewpy release, so dropping them without warning would cause serious headaches for users. I’d prefer adding `generate` as a new function; we might be able to turn (one of) the old functions into a thin layer on top of the new implementation, to avoid having too much duplicate code.
* I have similar worries about the other functions—making these changes without giving any warning by deprecating arguments first would break all existing code.
* Deprecating arguments in-place might work (but require some careful thought!); though before we go to that effort, I wonder whether it’s even worth preserving the three-step generate/simulate/collate structure of the API? In particular, the separation between `simulate` and `collate` comes from an era when `simulate` was done by calling the SNOwGLoBES binary and a separate step was necessary to read in the output files generated by SNOwGLoBES and make them available in snewpy. Now that collate is a fairly light post-processing step, maybe we could include it with the calculation step?

>  The detector_effects argument is now used - previously it was ignored.

Yep; this was left over from early snewpy versions, where we depended on (SNOw)GLoBES and transitioned to our internal rate calculation. We decided in #193 that it is no longer needed, since the performance overhead of calculating both unsmeared and smeared rates became sufficiently small. Unfortunately, it looks like we forgot to mark it as deprecated at the time. 🤦

jpkneller:
> ... it looks like you’re still working on it actively (with about 30 commits since you requested the review); so please let me know when it settles down and is ready for a thorough review.

The integration tests etc. and notebooks were giving me a lot of errors. I ended up having to touch a lot of files for small little edits.

> In the meantime, my first, high-level thoughts are:
> 
>     * We haven’t marked the `generate_*` functions as deprecated in any earlier snewpy release, so dropping them without warning would cause serious headaches for users. I’d prefer adding `generate` as a new function; we might be able to turn (one of) the old functions into a thin layer on top of the new implementation, to avoid having too much duplicate code.

Fair point. In the new version I retained the old **generate_* ** functions with an added warning that they are deprecated. One minor thing to note is that the new generate function **does not** integrate over the energy bins as the older **generate_* ** functions used to. In practice this should not cause a problem: the Container has a method called integrate_or_sum which will undertake the appropriate action for the circumstances, and by avoiding the energy bin integration in **generate**, the spectral fluences can be re-sampled by e,g, the RateCalculator if needed. 

>     * I have similar worries about the other functions—making these changes without giving any warning by deprecating arguments first would break all existing code.
>
>     * Deprecating arguments in-place might work (but require some careful thought!); though before we go to that effort, I wonder whether it’s even worth preserving the three-step generate/simulate/collate structure of the API? In particular, the separation between `simulate` and `collate` comes from an era when `simulate` was done by calling the SNOwGLoBES binary and a separate step was necessary to read in the output files generated by SNOwGLoBES and make them available in snewpy. Now that collate is a fairly light post-processing step, maybe we could include it with the calculation step?

The old versions of simulate and collate are also retained. The new function to replace them is called **calculate** and a) combines the effect of simulate and collate and b) returns Container objects. 

In a separate PR, I added a **collate** function to rate_calculator.py which might be useful. 

> > The detector_effects argument is now used - previously it was ignored.
> 
> Yep; this was left over from early snewpy versions, where we depended on (SNOw)GLoBES and transitioned to our internal rate calculation. We decided in #193 that it is no longer needed, since the performance overhead of calculating both unsmeared and smeared rates became sufficiently small. Unfortunately, it looks like we forgot to mark it as deprecated at the time. 🤦

I am open to having **calculate** add another level to the nested dictionary of event rate tables it returns, but is this really useful?

jpkneller:
I am going to keep this branch for reference but I am leaning more and more towards ditching this effort. If #445 is accepted, everything we need to easily replace the snowglobes module is available.

PR #445, opened 2026-07-07
Jpkneller add collate and aggregate functions for rate calculator tables

jpkneller:
Add collate function to process tables returned by RateCalculator.run


sybenzvi (commented):
I have no issues with the new `collate` calculation, but note that operator+ was newly defined for the flux containers. No issue with that either, but @Sheshuk should please weigh in.

jpkneller:
@sybenzvi, @Sheshuk, in the end I removed the overloading of the + operator for the Container because of the ambiguity in the flavor when adding two Containers with different flavors.

jpkneller:
The problems with the Detector_demo notebook are related to this PR because it too could use the + operator for Containers so I combined the two

jpkneller:
This PR also adds the __reduce__ and a _reconstruct method to Containers to allow dictionaries of Containers (i.e. the data structure returned by RateCalculator.run() ) to be saved to file with numpy's savez

PR #446, opened 2026-07-07
Add tests and notebooks for `Takata_2025` models

sybenzvi:
This PR started as a small effort to create a notebook to document the `Takata_2025` ALP production models, but in the course of that, I discovered the model constructor was broken in several places, so those got fixed as well.


sybenzvi:
Another time-critical bugfix.

PR #449, opened 2026-07-10
Update Yoshida constructor in presn_loaders.py

jpkneller:
eliminate double call to SupernovaModel constructor in preSN model Yoshida constructor


Sheshuk (approved):
Thanks for fixing this!

PR #452, opened 2026-07-13
Update extended.py

jpkneller:
Rewrite of extended.py so as to interface with rest of snewpy


sybenzvi:
@JostMigenda suggested during discussions on 2026-08-18 to see if we could add a way to check what part of a returned time series comes from the base model versus the extended tail. A few conditionals applied to `model.time` for the base model could allow us to do this within the class.

python/snewpy/models/extended.py
```diff
@@ -3,84 +3,78 @@
...
+            Times to evaluate luminosity.
 
         Returns
         -------
         astropy.Quantity
-            Luminosity calculated from cooling tail model.
+            extended time dependence calculated from cooling tail model.
         """
-        if t.value < 0.5:
+        if times[0] < 0.5*u.s:
```
sybenzvi, line 77: I don't love that we assume this is always an array, though it _should_ be given that the input arguments to `_get_initial_spectra_dict` are explicitly converted to arrays. Suggestion: use 
```
if np.asanyarray(times).item(0) < 0.5*u.s:
```

sybenzvi: On second thought, it's perhaps more robust to convert the input to an ndarray, which does so if `times` is a scalar and does nothing if it's already an array.

sybenzvi (approved):

PR #454, opened 2026-08-04
Update the SNEWPY --> MARLEY script for compatibility with MARLEY v2

sjgardiner:
I added an input parameter to specify the major version of MARLEY. It will default to v2 and use the new recommended physics configuration. The user can specify v1 instead as needed to fall back to the old script behavior.

Please test before merging! 😄


jpkneller (approved):
Looks good to me. It runs with no errors and the output looks to be what I would expect

JostMigenda:
Excellent; thanks for your contribution, @sjgardiner!

